PR #1, opened 2019-02-12
[NO SQUASH] New API

valhristov:
NOTE: it is better to review commit by commit


SonarQube.Client/Api/DefaultConfiguration.cs
```diff
@@ -39,6 +39,7 @@ public static RequestFactory Configure(RequestFactory requestFactory)
                 .RegisterRequest<IGetRoslynExportProfileRequest, V5_20.GetRoslynExportProfileRequest>("5.2")
                 .RegisterRequest<IGetModulesRequest, V5_40.GetModulesRequest>("5.4")
                 .RegisterRequest<IGetRulesRequest, V5_50.GetRulesRequest>("5.5")
+                .RegisterRequest<IDownloadStaticFile, V5_50.DownloadStaticFile>("5.5")
```
valhristov: The SonarQube request handler is actually implemented in the SonarC# plugin, so I guess the min supported version is older, but we don't support SQ before 5.6 (LTS-1) anyway...

SonarQube.Client/Api/V5_50/DownloadStaticFile.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    public class DownloadStaticFile : RequestBase<Stream>, IDownloadStaticFile
+    {
+        [JsonIgnore]
+        public string PluginKey { get; set; }
+
+        [JsonIgnore]
+        public string FileName { get; set; }
+
+        protected override string Path => $"static/{PluginKey}/{FileName}";
```
valhristov: The arguments are part of the path, hence they are not serialized as query string parameters.

SonarQube.Client.Tests/SonarQubeService_GetRulesAsync.cs
```diff
@@ -0,0 +1,219 @@
+﻿/*
+* SonarQube Client
+* Copyright (C) 2016-2018 SonarSource SA
```
duncanp-sonar: Should probably be "2016-2019". Still, almost all of the existing files need to be changed so I'll bulk update them  later.

SonarQube.Client/HttpResponseException.cs
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+using System;
+using System.Net;
+
+namespace SonarQube.Client
+{
+    public class HttpResponseException : Exception
+    {
+        public HttpResponseException(HttpStatusCode statusCode, string reasonPhrase)
```
duncanp-sonar: Exceptions should be serializable (there's a snippet that correctly implements the pattern - just type `exception` then hit `tab`, `tab`)

SonarQube.Client/Api/V6_30/GetPropertiesRequest.cs
```diff
@@ -61,8 +62,28 @@ public override async Task<SonarQubeProperty[]> InvokeAsync(HttpClient httpClien
...
+        private IEnumerable<SonarQubeProperty> ToProperties(PropertyResponse arg)
+        {
+            if (arg.FieldValues != null)
+            {
+                for (int i = 0; i < arg.FieldValues.Length; i++)
+                {
+                    var fieldValue = arg.FieldValues[i];
+                    foreach (var item in fieldValue)
+                    {
+                        yield return new SonarQubeProperty($"{arg.Key}.{i + 1}.{item.Key}", item.Value);
```
duncanp-sonar: One day, I'm going to profile this use of `yield`, and when the performance and memory allocations turn out to be much worse than just building and returning a list (not to mention making debugging harder), I'm going to write a rule to say "don't use yield like this" ;-)
Mind you, it could be difficult to separate out the false positives... ;-)

valhristov: The code is not performance critical, it should not matter much if you use yield or return array/list... :)

duncanp-sonar: True - it's just horrible to debug through ;-)

PR #4, opened 2019-03-21
Make sure the sonarqube service doesn't fail when reaching 10k projects

Evangelink:


SonarQube.Client/Requests/PagedRequestBase.cs
```diff
@@ -33,8 +33,8 @@ namespace SonarQube.Client.Requests
     /// <typeparam name="TResponseItem">The type of the items returned by this request.</typeparam>
     public abstract class PagedRequestBase<TResponseItem> : RequestBase<TResponseItem[]>, IPagedRequest<TResponseItem>
     {
-        private const int FirstPage = 1;
-        private const int MaximumPageSize = 500;
+        protected const int FirstPage = 1;
+        protected const int MaximumPageSize = 500;
```
duncanp-sonar: Do these need to be `protected`? They don't appear to be used outside this class.

Evangelink: This is no longer needed! I wasn't pleased with my first solution and forgot to revert this bit.

PR #5, opened 2019-03-22
Update solution and projects so that it can be properly referenced fr…

Evangelink:
…om SonarLint


SonarQube.Client.2015.sln
```diff
@@ -1,10 +1,12 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28307.329
+# Visual Studio 14
```
Evangelink: Make sure the 2015 solution is actually targeting VS2015.

SonarQube.Client.Tests/SonarQube.Client.Tests.2015.csproj
```diff
@@ -0,0 +1,143 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
Evangelink: Forge a VS2015 version of the test project that will be used by SonarLint for VS.

PR #7, opened 2019-07-12
Fix unit tests in non-English locale

henryju:
Fix for https://github.com/SonarSource/sonarlint-visualstudio/issues/902

### DIFF
--- a/SonarQube.Client.Tests/SonarQubeService_TestBase.cs
+++ b/SonarQube.Client.Tests/SonarQubeService_TestBase.cs
@@ -31,6 +31,7 @@ using Moq.Protected;
 using SonarQube.Client.Requests;
 using SonarQube.Client.Models;
 using SonarQube.Client.Api;
+using System.Globalization;
 
 namespace SonarQube.Client.Tests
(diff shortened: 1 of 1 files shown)


SonarQube.Client.Tests/SonarQubeService_TestBase.cs
```diff
@@ -49,6 +50,10 @@ public class SonarQubeService_TestBase
         [TestInitialize]
         public void TestInitialize()
         {
+            // Ensure exception messages are not platform dependent to not break assertions on non english platforms
+            Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
+            Thread.CurrentThread.CurrentUICulture = CultureInfo.InvariantCulture;
```
duncanp-sonar: I was wondering whether it would be necessary to reset the UI culture after the test has finished (in case the tests are run on a thread pool thread that is reused for other things). However, I've done a quick test running the tests inside VS and it looks like the test threads aren't reused, so I don't think resetting the culture is necessary.

PR #11, opened 2019-07-18
Changed service to offer HasOrganizations method

### LINKED ISSUE #10: Implement HasOrganizations method on ISonarQubeService
#### Description
The service has an "HasOrganizationsFeature" method that returns whether the API is available.
However, just because the API is available does not mean it is being used.

The service currently leaves it to the caller to correctly implement the logic of whether the organizations are actually being used or not. It would be better if the service offered a `HasOrganizations` feature method that correctly implemented the logic, instead of `HasOrganizationsFeature`.

duncanp-sonar:
Fixes #10


SonarQube.Client.Tests/SonarQubeService_HasOrganizations.cs
```diff
@@ -0,0 +1,189 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace SonarQube.Client.Tests
```
duncanp-sonar: I renamed the previous test class from `SonarQubeService_HasOrganizationFeature` (so the test class name matches the method name).
I copied across most of the old tests and amended them as appropriate.
As it happens, the old test class was missing the `[TestClass]` attribute so none of the old tests were being run.

SonarQube.Client/SonarQubeService.cs
```diff
@@ -47,14 +47,35 @@ public class SonarQubeService : ISonarQubeService, IDisposable
 
         public Version SonarQubeVersion { get; private set; }
 
-        public bool HasOrganizationsFeature
+        public async Task<bool> HasOrganizations(CancellationToken token)
```
duncanp-sonar: This is now a function call rather than a property because it does more than just a simple numerical check.

PR #42, opened 2020-04-23
Remove LangVersion settings from the projects

duncanp-sonar:
We'll rely on the default setting i.e. the compiler will warn us if we use a feature that is not supported
by the target framework.

For some reason the language level was set to a lower level than the one supported by the target framework.


SonarQube.Client.Tests/SonarQube.Client.Tests.csproj
```diff
@@ -7,7 +7,6 @@
     <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <RootNamespace>SonarQube.Client.Tests</RootNamespace>
-    <LangVersion>6</LangVersion>
```
duncanp-sonar, line 10: I can't think of any reason for setting this to restrict language features to a lower level than that supported by the target framework so I've removed it. It was preventing me from using pattern matching in the changes of the `RequestFactory`.

PR #43, opened 2020-04-23
Modified the factory to support requests that require the service

duncanp-sonar:


SonarQube.Client/Requests/RequestFactory.cs
```diff
@@ -98,10 +99,16 @@ public RequestFactory(ILogger logger)
         /// SonarQube version to return a request implementation for. When the provided value is null
         /// returns the registered implementation with the highest version number.
         /// </param>
+        /// <param name="service">The current SonarQubeService instance</param>
         /// <returns>New the newest TRequest implementation for the specified SonarQube version.</returns>
-        public TRequest Create<TRequest>(Version version)
+        public TRequest Create<TRequest>(Version version, ISonarQubeService service)
             where TRequest : IRequest
         {
+            if (service == null)
```
duncanp-sonar, line 107: Note: it's ok for the version to be null (the `param` doc comment above says that passing a null version selects the latest available version)

duncanp-sonar:
Doh! Failing test - I shouldn't have tried to simplify it...

SonarQube.Client/SonarQubeService.cs
```diff
@@ -110,7 +110,7 @@ public SonarQubeService(HttpMessageHandler messageHandler, RequestFactory reques
         {
             EnsureIsConnected();
 
-            var request = requestFactory.Create<TRequest>(SonarQubeVersion);
+            var request = requestFactory.Create<TRequest>(SonarQubeVersion, this);
```
rita-g-sonarsource, line 113: I see two inconsistencies:
1. Receiving the service in the request's ctor, means that all the requests will need to implement IDisposable, in order for the new request to dispose the service.
2. The signature of IRequest is InvokeAsync(httpClient, token), and the new request class is not gonna use httpClient.

I think we should standardize the way we pass parameters.
I can think of a few suggestions:

1. Add `Initialize(ISonarQubeService)` to `IRequest`, and have `ISonarQubeService` expose `HttpClient` property. This way IRequest.InvokeAsync will only receive Token, and each request will either use `ISonarQubeService.HttpClient` or `ISonarQubeService`.

2. Rename IRequiresSonarQubeService to `ICompositeRequest` with Initialize(ISonarQubeService) method, and create additional interface `IDirectRequest` with Initialize(HttpClient) method. Each request will either inherit from IDirectRequest or ICompositeRequest, and all will need to implement IDisposable.

Wdyt?

duncanp-sonar: > 1. Receiving the service in the request's ctor, means that all the requests will need to implement IDisposable, in order for the new request to dispose the service.

That doesn't follow - if an object is handed a disposable object, it doesn't mean that it becomes responsible for the lifetime of that object. In fact it's the opposite - you wouldn't expect the request (short-lived) to dispose the service (long-lived).

Exposing HttpClient on `ISonarQubeService` doesn't seem right either - consumers like SLVS shouldn't be able to see it.

I agree that adding the service to `IRequest` in all cases would be simpler.

How about the following?

```
public interface IRequest
{
    void Initialize(ISonarQubeService service, HttpClient httpClient, ILogger logger);
}

public interface IRequest<TResponse> : IRequest
{
    Task<TResponse> InvokeAsync(CancellationToken token);
}
```

... although we're still passing both ISonarQubeService and HttpClient to all requests, most of which will only use one of them. The only real difference is whether it's passed in `Initiailize` or `InvokeAsync`.

rita-g-sonarsource: Yep, ignore my comment regarding IDisposable.
I think it would be cleaner to have 2 different interfaces that inherit from IRequest, each with its own Initialize method. But we can start with adding all to IRequest and see how it goes next time we change it.

duncanp-sonar: Pass the service in `InvokeAsync` created less churn.

duncanp-sonar: I think adding more interfaces wouldn't add anything except complexity. From a practical point of view I don't think it matters if a request has access to both the `HttpClient` and the service (and it's possible to think of types of request that would need both).

PR #45, opened 2020-04-24
Added NuGePackageInfoGenerator

duncanp-sonar:
Adds a generator for the NuGetPackageInfo part of the RoslynExporter response.

### DIFF
--- a/SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
+++ b/SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
@@ -1,0 +1,59 @@
+﻿/*
+ * SonarQube Client
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.Collections.Generic;
+using System.Linq;
+using SonarQube.Client.Messages;
+using SonarQube.Client.Models;
+
+namespace SonarQube.Client.RoslynExporterAdapter
+{
+    public static class NuGetPackageInfoGenerator
+    {
+        public static IEnumerable<NuGetPackageInfoResponse> GetNuGetPackageInfos(IEnumerable<SonarQubeRule> activeRules,
+            IDictionary<string, string> sonarProperties)
+        {
+            var propertyPrefixes = GetPluginPropertyPrefixes(activeRules);
+            var packages = new List<NuGetPackageInfoResponse>();
+
+            foreach (var partialRepoKey in propertyPrefixes)
+            {
+                if (!sonarProperties.TryGetValue($"{partialRepoKey}.analyzerId", out var analyzerId) ||
+                    !sonarProperties.TryGetValue($"{partialRepoKey}.pluginVersion", out var pluginVersion))
+                {
+                    continue;
+                }
+
+                packages.Add(new NuGetPackageInfoResponse { Id = analyzerId, Version = pluginVersion });
+            }
+
+            return packages;
+        }
+
+        private static IEnumerable<string> GetPluginPropertyPrefixes(IEnumerable<SonarQubeRule> rules)
+        {
+            var partialRepoKeys = new HashSet<string>();
+            partialRepoKeys.UnionWith(rules.Select(r => r.TryGetRoslynPluginPropertyPrefix()));
+            partialRepoKeys.Remove(null);
+
+            return partialRepoKeys;
+        }
+    }
+}
--- a/SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
+++ b/SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
@@ -47,13 +47,10 @@ namespace SonarQube.Client.RoslynExporterAdapter
             return packages;
         }
 
-        private static IEnumerable<string> GetPluginPropertyPrefixes(IEnumerable<SonarQubeRule> rules)
-        {
-            var partialRepoKeys = new HashSet<string>();
-            partialRepoKeys.UnionWith(rules.Select(r => r.TryGetRoslynPluginPropertyPrefix()));
-            partialRepoKeys.Remove(null);
-
-            return partialRepoKeys;
-        }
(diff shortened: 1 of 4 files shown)


SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
```diff
@@ -0,0 +1,59 @@
...
+                packages.Add(new NuGetPackageInfoResponse { Id = analyzerId, Version = pluginVersion });
+            }
+
+            return packages;
+        }
+
+        private static IEnumerable<string> GetPluginPropertyPrefixes(IEnumerable<SonarQubeRule> rules)
+        {
+            var partialRepoKeys = new HashSet<string>();
+            partialRepoKeys.UnionWith(rules.Select(r => r.TryGetRoslynPluginPropertyPrefix()));
```
rita-g-sonarsource, line 53: I believe HashSet's ctor receives an enumerable to initialize

SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
```diff
@@ -0,0 +1,59 @@
...
+            partialRepoKeys.Remove(null);
```
rita-g-sonarsource, line 54: Why is there a null in the list?

SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
duncanp-sonar: It does, but it looks ugly and is harder to read.

SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
duncanp-sonar: `TryGet` returns `null` if the rule isn't for a Roslyn/SonarC/VB plugin.

SonarQube.Client/RoslynExporterAdapter/RoslynPluginRuleKeyExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                return VBNetPropertyPrefix;
+            }
+
+            // For plugins created by the SonarQube Roslyn SDK, the repository key is in the form
+            // "roslyn.MyCustomAnalyzer", and the Sonar properties added by the custom plugin will
+            // start with "MyCustomAnalyzer." e.g.
+            // * MyCustomAnalyzer.analyzerId
+            // * MyCustomAnalyzer.pluginVersion
+            if (rule.RepositoryKey.Length > ROSLYN_REPOSITORY_PREFIX.Length &&
```
rita-g-sonarsource, line 54: I think the Length check is a bit confusing to read, how about something like this:
if (rule.RepositoryKey.StartsWith(ROSLYN_REPOSITORY_PREFIX))
{
   var properyPrefix = rule.RepositoryKey.Substring(ROSLYN_REPOSITORY_PREFIX.Length);
return string.IsNullOrEmpty(properyPrefix) ? null : properyPrefix;
}
It should achieve the same thing but a bit easier to understand (IMO)

SonarQube.Client/RoslynExporterAdapter/RoslynPluginRuleKeyExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+                rule.RepositoryKey.StartsWith(ROSLYN_REPOSITORY_PREFIX))
+            {
+                return rule.RepositoryKey.Substring(ROSLYN_REPOSITORY_PREFIX.Length);
+            }
+
+            return null; // not a Roslyn-based rule
```
rita-g-sonarsource, line 60: When would we reach this point?
How do we have a rule that is not one of the 3 options?

SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+using System.Collections.Generic;
+using System.Linq;
+using SonarQube.Client.Messages;
+using SonarQube.Client.Models;
+
+namespace SonarQube.Client.RoslynExporterAdapter
+{
+    public static class NuGetPackageInfoGenerator
```
rita-g-sonarsource, line 28: I think Provider would be a more suitable name; Generator implies that there is some modification, rather than info collection.

SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
rita-g-sonarsource, line 54: The null is a bit confusing.
Also, maybe instead of a hashset, we can use Distinct(), and then Where(x=> !string.IsNullOrEmpty(x)

SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        public static IEnumerable<NuGetPackageInfoResponse> GetNuGetPackageInfos(IEnumerable<SonarQubeRule> activeRules,
+            IDictionary<string, string> sonarProperties)
+        {
+            var propertyPrefixes = GetPluginPropertyPrefixes(activeRules);
+            var packages = new List<NuGetPackageInfoResponse>();
+
+            foreach (var partialRepoKey in propertyPrefixes)
+            {
+                if (!sonarProperties.TryGetValue($"{partialRepoKey}.analyzerId", out var analyzerId) ||
```
rita-g-sonarsource, line 38: If we reached this point -- is it possible that those properties are missing?
It assumes that we already have a rule with `roslyn.<something>`, is it possible to have a rule without those properties?

rita-g-sonarsource:
@duncanp-sonar I'm not sure I understand the exact flow and use-cases, please see my comments.

SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
duncanp-sonar: It's called Generator for consistency with the existing RuleSet creator copied from S4MSB.
I think either name is ok; I don't agree that Generator implied modification, to me it means "creation".

SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
duncanp-sonar: It's a good practice to code defensively in extensibility scenarios i.e. where code we don't own is being plugged in. It's better to code defensively that crash at runtime because of what happens in third-party code.

However it shouldn't happen for plugins created using the SonarQube Roslyn SDK.

SonarQube.Client/RoslynExporterAdapter/NuGetPackageInfoGenerator.cs
duncanp-sonar: Ok, if you find it confusing,

SonarQube.Client/RoslynExporterAdapter/RoslynPluginRuleKeyExtensions.cs
duncanp-sonar: Ok, if you find it easier to read.

SonarQube.Client/RoslynExporterAdapter/RoslynPluginRuleKeyExtensions.cs
duncanp-sonar: Defensive coding again. so unlikely in practice. It could happen if a third-party wrote their own plugin for C# rules.

PR #46, opened 2020-04-27
Added SonarLint configuration generator

duncanp-sonar:


SonarQube.Client/RoslynExporterAdapter/Serializer.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Text;
+using System.Xml;
+using System.Xml.Serialization;
+
+namespace SonarQube.Client.RoslynExporterAdapter
+{
+    /// <summary>
+    /// Helper class to serialize objects to and from XML
+    /// </summary>
+    public static class Serializer
```
duncanp-sonar, line 32: There is a `Serializer` class in S4MSB, but it has more methods as it also handles loading and saving to disc.
There is a `ToString<T>` method in the S4MSB version, but it only seems to be called by test code (i.e. not used in production). Also, the S4MSB version doesn't correctly save the data as UTF-8.

The upshot is I haven't included a link to the S4MSB `Serializer` class since the implementations are sufficiently different.

SonarQube.Client/RoslynExporterAdapter/SonarLintConfigGenerator.cs
```diff
@@ -0,0 +1,108 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using SonarQube.Client.Models;
+
+// Logically equivalent to the SonarScanner for MSBuild class "RoslynSonarLint"
```
duncanp-sonar, line 26: This class should be functionally equivalent to the S4MSB version.
Unfortunately, the S4MSB implementation uses a `StringBuilder` to build the XML, so I didn't keep any of the implementation. Also, the S4MSB version doesn't have any tests, so I couldn't resuse those either.

rita-g-sonarsource: I see that the original class references "secured" properties, is that not relevant in our case?
https://github.com/SonarSource/sonar-scanner-msbuild/blob/9ccfdb648a0411014b29c7aee8e347aeab87ea71/src/SonarScanner.MSBuild.PreProcessor/Roslyn/RoslynSonarLint.cs#L46

SonarQube.Client/RoslynExporterAdapter/SonarLintConfigGenerator.cs
```diff
@@ -0,0 +1,108 @@
...
+            if (language == null) { throw new ArgumentNullException(nameof(language)); }
+
+            var slvsSettings = GetSettingsForLanguage(language, sonarProperties);
+
+            // We don't expect third-party rules to look for their settings in SonarLint.xml so we
+            // only fetch parameters for SonarC#/VB rules.
+            var sonarRepoKey = GetSonarRepoKey(language);
+            var slvsRules = GetRulesForRepo(sonarRepoKey, rules);
+
+            return new SonarLintConfiguration()
```
rita-g-sonarsource, line 50: What about the outer "AnalysisInput" node?
https://github.com/SonarSource/sonar-scanner-msbuild/blob/9ccfdb648a0411014b29c7aee8e347aeab87ea71/src/SonarScanner.MSBuild.PreProcessor/Roslyn/RoslynSonarLint.cs#L41

SonarQube.Client/RoslynExporterAdapter/SonarLintConfigGenerator.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            List<SonarLintKeyValuePair> slvsParameters = null;
+            if (sqRule.Parameters != null && sqRule.Parameters.Count > 0 )
+            {
+                slvsParameters = sqRule.Parameters.Select(ToSonarLintKeyValue).ToList();
+            }
+
+            return new SonarLintRule()
+            {
+                Key = sqRule.Key,
```
rita-g-sonarsource, line 100: There is a reference to "TemplateKey" in the original, is that not needed?
https://github.com/SonarSource/sonar-scanner-msbuild/blob/9ccfdb648a0411014b29c7aee8e347aeab87ea71/src/SonarScanner.MSBuild.PreProcessor/Roslyn/RoslynSonarLint.cs#L58

SonarQube.Client/RoslynExporterAdapter/SonarLintConfigGenerator.cs
duncanp-sonar: No, I don't they exist any more. The secured properties were for the C++ and VB net plugins when they were commerically licensed.

SonarQube.Client/RoslynExporterAdapter/SonarLintConfigGenerator.cs
duncanp-sonar: It doesn't exist as a field in the `SonarQubeRule` data class used by the client library, so it's never been used by SonarLint.

FYI I think the TemplateKey relates to "templated rules" - a simple but limited way of allowing users to add their custom rules. However, they are not supported by all languages (including C#, VB and C++).

PR #48, opened 2020-05-12
Added issue type to SonarQubeRule

duncanp-sonar:
Required by SLVS to filter out hotspots.

### DIFF
--- a/SonarQube.Client.Tests/Helpers/SonarQubeIssueTypeConverterTests.cs
+++ b/SonarQube.Client.Tests/Helpers/SonarQubeIssueTypeConverterTests.cs
@@ -1,0 +1,51 @@
+﻿/*
+ * SonarQube Client
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using FluentAssertions;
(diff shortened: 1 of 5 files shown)


SonarQube.Client.Tests/Helpers/SonarQubeIssueTypeConverterTests.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        [TestMethod]
+        // Exact matches
+        [DataRow("CODE_SMELL", SonarQubeIssueType.CodeSmell)]
+        [DataRow("BUG", SonarQubeIssueType.Bug)]
+        [DataRow("VULNERABILITY", SonarQubeIssueType.Vulnerability)]
+        [DataRow("SECURITY_HOTSPOT", SonarQubeIssueType.SecurityHotspot)]
+
+        // Case-insensitivity
+        [DataRow("bug", SonarQubeIssueType.Bug)]
```
duncanp-sonar, line 38: I'm not sure we really care about allowing case-insensitivity, but the existing `SonarQubeIssueSeverity` converter does so I've done the same here.

SonarQube.Client.Tests/SonarQubeService_GetRulesAsync.cs
```diff
@@ -135,6 +135,7 @@ public async Task GetRulesAsync_Active_SonarQubeResponse()
             result.Select(r => r.RepositoryKey).Should().Contain(new[] { "csharpsquid", "csharpsquid", "csharpsquid" });
             result.Select(r => r.IsActive).Should().Contain(new[] { true, true, true });
             result.Select(r => r.Severity).Should().ContainInOrder(new[] { SonarQubeIssueSeverity.Major, SonarQubeIssueSeverity.Critical, SonarQubeIssueSeverity.Minor });
+            result.Select(r => r.IssueType).Should().ContainInOrder(new[] { SonarQubeIssueType.Bug, SonarQubeIssueType.Vulnerability, SonarQubeIssueType.CodeSmell });
```
duncanp-sonar, line 138: The json response in the test already had a type information in it, so I've just had to add the assertion.

SonarQube.Client/Api/V5_50/GetRulesRequest.cs
```diff
@@ -99,6 +101,9 @@ private class RuleResponse
 
             [JsonProperty("repo")]
             public string RepositoryKey { get; set; }
+
+            [JsonProperty("type")]
+            public string Type { get; set; }
```
duncanp-sonar, line 106: Fortunately, there is only one version of `GetRulesRequest` i.e. the API hasn't changed since this version was created.
Even better, I didn't need to change the request to include additional data since the returned json already contained the type information.
All I needed to do was deserialize the json and convert it to an enum.

PR #49, opened 2020-05-25
Fix paging to avoid error when reaching 10k API limit

### LINKED ISSUE #8: Paging is not handled correctly
The `PagedRequestBase` requests additional pages of data until a partial page of results is returned. What is considered a partial page of results should be determined by `request.PageSize`. 

However, the class always compares the number of results against the maximum allowed number results per page (500).

The current incorrect behaviour is relied on by some of the requests e.g. `GetQualityProfile`. It only wants a single result, so it sets the request page size to 1. The server only returns one result, and the paged base class doesn't request any more pages since the number of results returned is less than 500.

#### Miscellaneous - c

rita-g-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/1292
Fixes #8

### DIFF
--- a/SonarQube.Client/Api/V7_20/GetIssuesRequestWrapper.cs
+++ b/SonarQube.Client/Api/V7_20/GetIssuesRequestWrapper.cs
@@ -60,22 +60,22 @@ namespace SonarQube.Client.Api.V7_20
             innerRequest.ProjectKey = ProjectKey;
             innerRequest.Statuses = Statuses;
             innerRequest.Logger = Logger;
-            innerRequest.MaxPageNumber = MaxNumberOfSupportedIssues / MaxNumberOfIssuesPerPage + 1;
+            innerRequest.MaxPageNumber = MaxNumberOfSupportedIssues / MaxNumberOfIssuesPerPage;
 
             ResetInnerRequest();
             innerRequest.Types = "CODE_SMELL";
             var codeSmells = await innerRequest.InvokeAsync(httpClient, token);
-            WarnForApiLimit(codeSmells, innerRequest);
+            WarnForApiLimit(codeSmells);
 
             ResetInnerRequest();
             innerRequest.Types = "BUG";
             var bugs = await innerRequest.InvokeAsync(httpClient, token);
-            WarnForApiLimit(bugs, innerRequest);
+            WarnForApiLimit(bugs);
 
             ResetInnerRequest();
             innerRequest.Types = "VULNERABILITY";
             var vulnerabilities = await innerRequest.InvokeAsync(httpClient, token);
-            WarnForApiLimit(vulnerabilities, innerRequest);
+            WarnForApiLimit(vulnerabilities);
 
             return codeSmells
                 .Concat(bugs)
@@ -83,11 +83,11 @@ namespace SonarQube.Client.Api.V7_20
                 .ToArray();
         }
 
-        private void WarnForApiLimit(SonarQubeIssue[] issues, GetIssuesRequest request)
+        private void WarnForApiLimit(SonarQubeIssue[] issues)
         {
-            if (issues.Length == MaxNumberOfSupportedIssues || request.Page == request.MaxPageNumber)
+            if (issues.Length == MaxNumberOfSupportedIssues)
             {
-                Logger.Warning($"The SonarQube maximum API response limit reached. Some issues might not be suppressed.");
+                Logger.Warning($"The SonarQube maximum API response limit reached. Some issues might not be suppressed, suppressing the first {MaxNumberOfSupportedIssues} issues.");
             }
         }
 
--- a/SonarQube.Client/Api/V7_20/GetIssuesRequestWrapper.cs
+++ b/SonarQube.Client/Api/V7_20/GetIssuesRequestWrapper.cs
@@ -59,17 +59,17 @@ namespace SonarQube.Client.Api.V7_20
             ResetInnerRequest();
             innerRequest.Types = "CODE_SMELL";
             var codeSmells = await innerRequest.InvokeAsync(httpClient, token);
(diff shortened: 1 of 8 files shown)


SonarQube.Client/Requests/PagedRequestBase.cs
```diff
@@ -60,7 +63,8 @@ public async override Task<TResponseItem[]> InvokeAsync(HttpClient httpClient, C
 
                 Page++;
             }
-            while (pageResult.Value != null &&
+            while (!ReachedMaxPageNumber() &&
+                pageResult.Value != null &&
```
rita-g-sonarsource, line 67: @duncanp-sonar do you want to fix https://github.com/SonarSource/sonarqube-webclient-dotnet/issues/8 as well?

SonarQube.Client/Requests/PagedRequestBase.cs
```diff
@@ -69,6 +73,11 @@ public async override Task<TResponseItem[]> InvokeAsync(HttpClient httpClient, C
             return allResponseItems.ToArray();
         }
 
+        private bool ReachedMaxPageNumber()
+        {
+            return MaxPageNumber.HasValue && Page > MaxPageNumber.Value;
```
duncanp-sonar, line 78: Is the problem the maximum number of pages or the maximum number of results?

Do any other APIs have this limitation on the maximum number of results?

SonarQube.Client/Requests/PagedRequestBase.cs
duncanp-sonar: As discussed offline, changing this check to be "max issues reached" might fix the other issue too, but we'd need to check that the behaviour of e.g. `GetQualityProfile` isn't broken.

SonarQube.Client/Api/V7_20/GetIssuesRequestWrapper.cs
```diff
@@ -56,6 +57,7 @@ public async Task<SonarQubeIssue[]> InvokeAsync(HttpClient httpClient, Cancellat
             innerRequest.ProjectKey = ProjectKey;
             innerRequest.Statuses = Statuses;
             innerRequest.Logger = Logger;
+            innerRequest.MaxPageNumber = 20; // Retrieve 10000 in 20 pages x 500 items per page;
```
duncanp-sonar, line 60: The way the client works means if the server version is less than v7.2 this request class won't be used, so the bug will still occur.

The earliest supported `GetIssues` request API is for v5.1. Technically earlier versions of SQ are out of support. However, if there is no additional cost for us in fixing the issue for earlier versions then we might as well do it: it helps customers, and it might also simplify support issues for us in the future (i.e. we won't have to remember that we only fixed this bug for some versions of SQ).

SonarQube.Client/Requests/PagedRequestBase.cs
```diff
@@ -60,13 +65,24 @@ public async override Task<TResponseItem[]> InvokeAsync(HttpClient httpClient, C
 
                 Page++;
             }
-            while (pageResult.Value != null &&
-                // Continue paging until we get a partial page of results i.e. fewer than requested.
-                // NB there is a bug here: should be comparing against the request page size, not the
-                // maximum allowed size. See https://github.com/SonarSource/sonarqube-webclient-dotnet/issues/8
-                pageResult.Value.Length >= MaximumPageSize);
+            while (!ReachedMaxPageNumber() &&
```
duncanp-sonar, line 68: This seems unnecessarily complicated. Why not just check `allResponseItems.Count` against the limit?

SonarQube.Client/Api/V7_20/GetIssuesRequestWrapper.cs
```diff
@@ -60,29 +59,29 @@ public async Task<SonarQubeIssue[]> InvokeAsync(HttpClient httpClient, Cancellat
...
         }
 
-        private void WarnForApiLimit(SonarQubeIssue[] issues)
+        private void WarnForApiLimit(SonarQubeIssue[] issues, GetIssuesRequest request)
         {
-            if (issues.Length == 10000)
+            if (issues.Length == request.ItemsLimit)
             {
-                Logger.Warning($"The SonarQube maximum API response limit reached. Some issues might not be suppressed.");
+                Logger.Warning($"The SonarQube maximum API response limit reached. Some issues might not be suppressed, suppressing the first {request.ItemsLimit} issues.");
```
duncanp-sonar, line 84: The sentence is a bit clumsy and I don't think it's grammatically correct (in fact, the original sentence wasn't grammatically correct either). It might be better to not refer to SonarQube specifically since SonarCloud users will see the same message. Here a couple of alternatives:

>Sonar web API response limit reached ({request.ItemsLimit} items). Some issues might not be suppressed.

or 

>Sonar web maximum API response limit reached. Only the first {request.ItemsLimit} issues will be suppressed.

SonarQube.Client/Requests/PagedRequestBase.cs
rita-g-sonarsource: Oh wow, how did I not see it. Tunnel-vision!

SonarQube.Client/Requests/PagedRequestBase.cs
```diff
@@ -60,13 +65,24 @@ public async override Task<TResponseItem[]> InvokeAsync(HttpClient httpClient, C
...
+                pageResult.Value != null &&
+                pageResult.Value.Length >= PageSize);
+
+            return allResponseItems.Take(ItemsLimit).ToArray();
```
duncanp-sonar, line 72: This will be redundant in 99.99% of cases, and it's not immediately obvious why it is there.
I think it would be clearer and more obvious if this were only done if there are more items than requested.

PR #54, opened 2020-06-08
Fixed incorrectly-displayed "max items retrieved" message

duncanp-sonar:
Fixed #1450

### DIFF
--- a/SonarQube.Client/Requests/PagedRequestBase.cs
+++ b/SonarQube.Client/Requests/PagedRequestBase.cs
@@ -19,6 +19,7 @@
  */
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Net.Http;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


SonarQube.Client/Requests/PagedRequestBase.cs
```diff
@@ -73,10 +73,12 @@ public override async Task<TResponseItem[]> InvokeAsync(HttpClient httpClient, C
                 return allResponseItems.ToArray();
             }
 
-            Logger.Warning("Sonar web maximum API response limit reached.");
+            if (allResponseItems.Count == MaximumItemsCount)
```
duncanp-sonar, line 76: We should probably validate that caller hasn't set the `ItemsLimit` to a number greater than `MaximumItemsCount`, but that's a separate problem.

duncanp-sonar: It's also difficult to write a test for without refactoring, so I'll add an assertion for now.

PR #55, opened 2020-08-13
Upgrade to Moq 4.14.5

rita-g-sonarsource:


SonarQube.Client.Tests/SonarQubeService_Lifecycle.cs
```diff
@@ -86,14 +87,16 @@ public async Task Disconnect_Does_Not_Dispose_MessageHandler()
...
+            disposed.Should().BeFalse();
         }
 
         [TestMethod]
         public async Task Dispose_Does_Dispose_MessageHandler()
         {
             // Arrange
-            messageHandler.Protected().Setup("Dispose", true);
+            var disposed = false;
+            messageHandler.Protected().Setup("Dispose", ItExpr.IsAny<bool>()).Callback(() => disposed = true);
```
rita-g-sonarsource, line 98: The new version of moq complains 
```
Method HttpMessageHandler.Dispose is public. Use strong-typed Expect overload instead:
mock.Setup(x => x.Dispose());
```
But Dispose is not virtual, so you can't mock it.

So they [suggest](https://github.com/moq/moq4/issues/129#issuecomment-57915712) to use `mock.As<IDisposable>().Setup(s => s.Dispose());`, but then you can't verify it, since the code calls `HttpMessageHandler.Dispose` and not `(HttpMessageHandler as IDisposable).Dispose`.
So this is an ugly workaround.

PR #57, opened 2020-10-09
Added flow information to SonarQubeIssue

duncanp-sonar:
Fixes SLVS #1761

### DIFF
--- a/SonarQube.Client.Tests/Infra/MocksHelper.cs
+++ b/SonarQube.Client.Tests/Infra/MocksHelper.cs
@@ -1,0 +1,51 @@
+﻿/*
+ * SonarQube Client
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Moq;
+using Moq.Protected;
+
+namespace SonarQube.Client.Tests.Infra
+{
+    internal static class MocksHelper
+    {
+        public const string ValidBaseAddress = "http://localhost";
+
+        public static void SetupHttpRequest(Mock<HttpMessageHandler> messageHandlerMock, string requestRelativePath, string response,
+            HttpStatusCode statusCode = HttpStatusCode.OK, string basePath = ValidBaseAddress)
+        {
+            var baseUri = new Uri(basePath);
+            messageHandlerMock.Protected()
+                .Setup<Task<HttpResponseMessage>>("SendAsync",
+                    ItExpr.Is<HttpRequestMessage>(m =>
+                        m.RequestUri == new Uri(baseUri, requestRelativePath)),
+                    ItExpr.IsAny<CancellationToken>())
+                .Returns(Task.FromResult(new HttpResponseMessage
+                {
+                    StatusCode = statusCode,
+                    Content = new StringContent(response)
+                }));
+        }
+    }
+}
--- a/SonarQube.Client.Tests/Models/IssueFlowTests.cs
+++ b/SonarQube.Client.Tests/Models/IssueFlowTests.cs
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+using System.Collections.Generic;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarQube.Client.Models;
@@ -32,7 +33,21 @@ namespace SonarQube.Client.Tests.Models
         {
             var testSubject = new IssueFlow(null);
 
-            testSubject.Locations.Should().NotBeNull();
+            testSubject.Locations.Should().BeEmpty();
+        }
+
+        [TestMethod]
+        public void Ctor_PropertiesAreSet()
+        {
+            var locations = new List<IssueLocation>
+            {
+                new IssueLocation("component1", null, "message1"),
+                new IssueLocation("component2", null, "message2")
+            };
+
+            var testSubject = new IssueFlow(locations);
+
+            testSubject.Locations.Should().BeEquivalentTo(locations[0], locations[1]);
         }
     }
 }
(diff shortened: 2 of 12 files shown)


SonarQube.Client.Tests/Infra/MocksHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+using Moq;
+using Moq.Protected;
+
+namespace SonarQube.Client.Tests.Infra
+{
+    internal static class MocksHelper
+    {
+        public const string ValidBaseAddress = "http://localhost";
+
+        public static void SetupHttpRequest(Mock<HttpMessageHandler> messageHandlerMock, string requestRelativePath, string response,
```
duncanp-sonar, line 35: De-duplicating mock setup code

SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestTests.cs
```diff
@@ -0,0 +1,233 @@
...
+using SonarQube.Client.Api.V7_20;
+using static SonarQube.Client.Tests.Infra.MocksHelper;
+
+namespace SonarQube.Client.Tests.Requests.Api.V7_20
+{
+    [TestClass]
+    public class GetIssuesRequestTests
+    {
+        [TestMethod]
+        public async Task Invoke()
```
duncanp-sonar, line 38: Previously the GetIssues class was only tested indirectly. I've added a single direct test to cover the processing of the response and the conversion from json to the public data classes.

SonarQube.Client/Api/V5_10/GetIssuesRequest.cs
```diff
@@ -88,6 +88,6 @@ private static IEnumerable<T> ReadFromProtobufStream<T>(Stream stream, MessagePa
 
         private static SonarQubeIssue ToSonarQubeIssue(ServerIssue issue) =>
             new SonarQubeIssue(issue.Path, issue.Checksum, issue.Line, issue.Msg, issue.ModuleKey, issue.RuleKey,
-                issue.Status == "RESOLVED");
+                issue.Status == "RESOLVED", flows: null);
```
duncanp-sonar, line 91: This request class handles very old SQ instances. In that version the response was returned as a serialized protobuf stream. I tried installing a SQ5.1 instance so I could generate the expected protobuf response for the test, but that version of SQ is so old that it needs JRE v2 so I decided it wasn't worth it.

Instead, I added a test at the model level to make sure the flows collection is never null in the returned read-only data classes.

SonarQube.Client.Tests/Models/SonarQubeIssueTests.cs
```diff
@@ -0,0 +1,55 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarQube.Client.Models;
+
+namespace SonarQube.Client.Tests.Models
+{
+    [TestClass]
+    public class SonarQubeIssueTests
+    {
+        [TestMethod]
+        public void Ctor_FilePathCanBeNull()
```
duncanp-sonar, line 32: Tests to show/check null handling. The first two tests are for pre-existing behaviours, the third checks that flows are never null.

SonarQube.Client/Api/V5_10/GetIssuesRequest.cs
duncanp-sonar, line 91: I don't known why the returned status is hard-coded to `RESOLVED` - possibly because we don't use it at the moment. I'll create a separate ticket to look at it.

SonarQube.Client/Models/SonarQubeIssue.cs
```diff
@@ -41,5 +45,48 @@ public class SonarQubeIssue
...
+    public class IssueLocation
+    {
+        public IssueLocation(string component, IssueTextRange textRange, string message)
+        {
+            Component = component;
+            TextRange = textRange;
+            Message = message;
+        }
+
+        public string Component { get; }
```
rita-g-sonarsource, line 72: Is `Component` = FilePath? `SonarQubeIssue` has a FilePath property, so I'd expect that to be the naming convention.

SonarQube.Client.Tests/Models/IssueFlowTests.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarQube.Client.Models;
+
+namespace SonarQube.Client.Tests.Models
+{
+    [TestClass]
+    public class IssueFlowTests
+    {
+        [TestMethod]
+        public void Ctor_LocationsAreNeverNull()
```
rita-g-sonarsource, line 31: Might be useful to also have a test that checks that the passed locations are used.

SonarQube.Client.Tests/Models/SonarQubeIssueTests.cs
rita-g-sonarsource: I think it's worth having not-null tests as well, e.g. that the passed flows are used.

SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestTests.cs
```diff
@@ -0,0 +1,233 @@
...
+
+            var location_0_0_0 = result[0].Flows[0].Locations[0];
+            location_0_0_0.Component.Should().Be("myprojectkey:projectroot/Controllers/WeatherForecastController.cs");
+            location_0_0_0.Message.Should().Be("sink: tainted value is used to perform a security-sensitive operation");
+            location_0_0_0.TextRange.StartLine.Should().Be(43);
+            location_0_0_0.TextRange.EndLine.Should().Be(43);
+            location_0_0_0.TextRange.StartOffset.Should().Be(19);
+            location_0_0_0.TextRange.EndOffset.Should().Be(43);
+
+            var location_0_1_1 = result[0].Flows[1].Locations[1];
```
rita-g-sonarsource, line 214: I think the name is complicated to read. 
Since there is only 1 issue, you can extract `result[0]` into a variable, and then name these `firstFlowLocation` and `secondFlowLocation` or something similar. I think it would help with readability.

rita-g-sonarsource, line 214: What happened to location_0_1_0 ?

rita-g-sonarsource (approved):
LGTM, left some comments

SonarQube.Client/Models/SonarQubeIssue.cs
duncanp-sonar: Good catch.

Not exactly. Here,  `Component` is a composite data item in the form `[project key]:[relative file path]`. The code to create SonarQubeIssue instances has specific code to handle extracting the file path part. I'll create a separate PR for that.

SonarQube.Client.Tests/Models/IssueFlowTests.cs
duncanp-sonar: Ok, but it seems a bit of a junk test.

SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestTests.cs
duncanp-sonar: The main cases I want to test are:
1) properties for a single location are set properly,
2) multiple flows are handled, and
3) multiple locations are handled.
Testing a subset of the locations is enough to do that; I don't think testing every location adds much value.

SonarQube.Client.Tests/Models/SonarQubeIssueTests.cs
duncanp-sonar: Ok but I think it's very low value.

PR #61, opened 2020-10-15
Added GetHotspot method to API

duncanp-sonar:
Relates to https://github.com/SonarSource/sonarlint-visualstudio/issues/1771

### DIFF
--- a/SonarQube.Client.Tests/Requests/DefaultConfiguration_Configure.cs
+++ b/SonarQube.Client.Tests/Requests/DefaultConfiguration_Configure.cs
@@ -23,6 +23,7 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 using SonarQube.Client.Requests;
 using SonarQube.Client.Api;
 using SonarQube.Client.Tests.Infra;
+using System.Diagnostics;
 
 namespace SonarQube.Client.Tests.Requests
 {
@@ -34,10 +35,7 @@ namespace SonarQube.Client.Tests.Requests
         {
             var logger = new TestLogger();
 
-            DefaultConfiguration.Configure(new RequestFactory(logger));
-
-            logger.DebugMessages.Should().ContainInOrder(
-                new[]
+            var expected = new string[]
                 {
                     "Registered SonarQube.Client.Api.V2_10.GetPluginsRequest for 2.1",
                     "Registered SonarQube.Client.Api.V2_10.GetProjectsRequest for 2.1",
@@ -45,10 +43,13 @@ namespace SonarQube.Client.Tests.Requests
                     "Registered SonarQube.Client.Api.V2_60.GetPropertiesRequest for 2.6",
                     "Registered SonarQube.Client.Api.V3_30.ValidateCredentialsRequest for 3.3",
                     "Registered SonarQube.Client.Api.V5_10.GetIssuesRequest for 5.1",
+                    "Registered SonarQube.Client.Api.V5_10.GetLanguagesRequest for 5.1",
                     "Registered SonarQube.Client.Api.V5_20.GetQualityProfileChangeLogRequest for 5.2",
                     "Registered SonarQube.Client.Api.V5_20.GetQualityProfilesRequest for 5.2",
                     "Registered SonarQube.Client.Api.V5_20.GetRoslynExportProfileRequest for 5.2",
                     "Registered SonarQube.Client.Api.V5_40.GetModulesRequest for 5.4",
+                    "Registered SonarQube.Client.Api.V5_50.GetRulesRequest for 5.5",
+                    "Registered SonarQube.Client.Api.V5_50.DownloadStaticFile for 5.5",
                     "Registered SonarQube.Client.Api.V6_20.GetOrganizationsRequest for 6.2",
                     "Registered SonarQube.Client.Api.V6_20.GetProjectsRequest for 6.2",
                     "Registered SonarQube.Client.Api.V6_30.GetPluginsRequest for 6.3",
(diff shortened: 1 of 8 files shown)


SonarQube.Client.Tests/Requests/DefaultConfiguration_Configure.cs
```diff
@@ -34,21 +35,21 @@ public void Configure_Writes_Debug_Messages()
         {
             var logger = new TestLogger();
 
-            DefaultConfiguration.Configure(new RequestFactory(logger));
-
-            logger.DebugMessages.Should().ContainInOrder(
-                new[]
+            var expected = new string[]
```
duncanp-sonar, line 38: Previously the test didn't fail if there were unexpected requests in the list. Now it does.

SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
```diff
@@ -0,0 +1,185 @@
...
+            result.ComponentPath.Should().Be("src/main/java/com/sonarsource/FourthClass.java");
+
+            result.RuleKey.Should().Be("java:S4787");
+            result.RuleName.Should().Be("rule-name");
+            result.SecurityCategory.Should().Be("others");
+            result.VulnerabilityProbability.Should().Be("LOW");
+        }
+
+        [TestMethod]
+        public void GetHotspotAsync_NotConnected()
```
duncanp-sonar, line 171: This is a boilerplate test that exists for every public API.

SonarQube.Client/Api/V8_1/GetHotspotRequest.cs
```diff
@@ -0,0 +1,122 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using SonarQube.Client.Models;
+using SonarQube.Client.Requests;
+
+namespace SonarQube.Client.Api.V8_1
+{
+    public class GetHotspotRequest : RequestBase<SonarQubeHotspot>, IGetHotspotRequest
+    {
+        //TODO - this is an internal API - change to use a public API when available
```
duncanp-sonar, line 30: I didn't worry too much about the structure of the data that is being returned as I expect it will change in the real API.
For now we just need something that works end-to-end with the currently available APIs, and an interface we can mock in SLVS tests.

rita-g-sonarsource: Are we sure that the API will change? Perhaps we should wait with the merge to master, and instead work on a branch, like with did in the prototype. WDYT?

duncanp-sonar: The current API is internal and doesn't return secondary location information, so it seems pretty likely the API will change.
The PR is targeting a branch called `branch-taint`.

SonarQube.Client/Api/DefaultConfiguration.cs
```diff
@@ -50,7 +50,7 @@ public static RequestFactory Configure(RequestFactory requestFactory)
                 .RegisterRequest<IGetRoslynExportProfileRequest, V6_60.GetRoslynExportProfileRequest>("6.6")
                 .RegisterRequest<IGetOrganizationsRequest, V7_00.GetOrganizationsRequest>("7.0")
                 .RegisterRequest<IGetIssuesRequest, V7_20.GetIssuesRequestWrapper>("7.2")
-                ;
+                .RegisterRequest<IGetHotspotRequest, V8_1.GetHotspotRequest>("8.1");
```
rita-g-sonarsource, line 53: What should we do for older SQ versions? Perhaps we should have some kind of message in the hotspots window. WDYT?

SonarQube.Client.Tests/Requests/DefaultConfiguration_Configure.cs
```diff
@@ -59,7 +60,24 @@ public void Configure_Writes_Debug_Messages()
...
+            logger.DebugMessages.Should().ContainInOrder(expected);
+            logger.DebugMessages.Count.Should().Be(expected.Length);
+        }
+
+        private static void DumpDebugMessages(TestLogger logger)
+        {
+            Debug.WriteLine("Actual registered requests:");
+            foreach (var message in logger.DebugMessages)
+            {
+                Debug.WriteLine(message);
```
rita-g-sonarsource, line 79: There is a logger in this assembly, no? Can we use it?

SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
```diff
@@ -0,0 +1,185 @@
...
+            result.Assignee.Should().Be("joe");
+            result.Status.Should().Be("TO_REVIEW");
+            result.Line.Should().Be(10);
+
+            result.Organization.Should().Be("default-organization");
+            result.ProjectKey.Should().Be("com.sonarsource:test-project");
+            result.ProjectName.Should().Be("test-project");
+
+            result.ComponentKey.Should().Be("com.sonarsource:test-project:src/main/java/com/sonarsource/FourthClass.java");
+            result.ComponentPath.Should().Be("src/main/java/com/sonarsource/FourthClass.java");
```
rita-g-sonarsource, line 162: Perhaps ComponentPath should be FilePath, don't know if it's better to stay loyal to the returned values, or to make it more explicit for us to understand.

rita-g-sonarsource (approved):

SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
duncanp-sonar: Since I'm expecting the API to change I've done the absolute minimum necessary now to give us something we can consume in the client. I didn't think it was worth doing any extra work at this point.

SonarQube.Client/Api/DefaultConfiguration.cs
duncanp-sonar: (discussed offline, note added to the project board)

PR #62, opened 2020-11-03
Add TextRange to hotspots

rita-g-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/1858

### DIFF
--- a/SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
+++ b/SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
@@ -60,8 +60,13 @@ namespace SonarQube.Client.Tests
     ""securityCategory"": ""others"",
     ""vulnerabilityProbability"": ""LOW""
   },
+  ""textRange"": {
+    ""startLine"": 5,
+    ""endLine"": 10,
+    ""startOffset"": 29,
+    ""endOffset"": 41
+  },
   ""status"": ""TO_REVIEW"",
-  ""line"": 10,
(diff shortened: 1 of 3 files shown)


SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
```diff
@@ -60,8 +60,13 @@ public async Task GetHotspot_Response_From_SonarQube()
     ""securityCategory"": ""others"",
     ""vulnerabilityProbability"": ""LOW""
   },
+  ""textRange"": {
```
duncanp-sonar, line 63: Hhmm; interesting. This element doesn't appear in the example response in the webapi documentation for either 'next' (v8.6) or 'peach' (v8.5) (which is where the example used in the test came from). Is it just the webapi documentation that is wrong?

Also, assuming that the product code does now return the textRange, do we need to update the minimum version in our client code in which this API is supported? (currently listed as v8.1).

duncanp-sonar (commented):
The changes look ok; just a question about when the textRange started being supported in the response.

SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
rita-g-sonarsource: Good question; I took the response directly from the network, so that's what's being returned from Next at the moment.
I'll ask SQ folks regarding the docs + version.

PR #63, opened 2020-11-10
Normalize hotspot and issue file paths

rita-g-sonarsource:
Part of https://github.com/SonarSource/sonarlint-visualstudio/issues/1763

### DIFF
--- a/SonarQube.Client/Api/V5_10/GetIssuesRequest.cs
+++ b/SonarQube.Client/Api/V5_10/GetIssuesRequest.cs
@@ -27,6 +27,7 @@ using System.Net.Http;
 using System.Threading.Tasks;
 using Google.Protobuf;
 using Newtonsoft.Json;
+using SonarQube.Client.Helpers;
 using SonarQube.Client.Messages;
 using SonarQube.Client.Models;
 using SonarQube.Client.Requests;
@@ -87,7 +88,7 @@ namespace SonarQube.Client.Api.V5_10
         }
 
         private static SonarQubeIssue ToSonarQubeIssue(ServerIssue issue) =>
-            new SonarQubeIssue(issue.Path, issue.Checksum, issue.Line, issue.Msg, issue.ModuleKey, issue.RuleKey,
+            new SonarQubeIssue(FilePathNormalizer.NormalizeSonarQubePath(issue.Path), issue.Checksum, issue.Line, issue.Msg, issue.ModuleKey, issue.RuleKey,
                 issue.Status == "RESOLVED", flows: null);
     }
 }
--- a/SonarQube.Client/Api/V7_20/GetIssuesRequest.cs
+++ b/SonarQube.Client/Api/V7_20/GetIssuesRequest.cs
@@ -23,6 +23,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using SonarQube.Client.Helpers;
 using SonarQube.Client.Models;
 using SonarQube.Client.Requests;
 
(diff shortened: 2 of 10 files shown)


SonarQube.Client/Models/SonarQubeIssue.cs
```diff
@@ -28,7 +29,7 @@ public class SonarQubeIssue
         public SonarQubeIssue(string filePath, string hash, int? line, string message, string moduleKey, string ruleId,
             bool isResolved, List<IssueFlow> flows)
         {
-            FilePath = filePath?.Trim('/', '\\');
+            FilePath = filePath;
```
rita-g-sonarsource, line 32: I moved it to GetRequest classes. 
It's already responsible for handle data changes, such as calculating `isResolved`, so it made sense that the file path fixes will there as well.

SonarQube.Client/Api/V5_10/GetIssuesRequest.cs
```diff
@@ -87,7 +88,7 @@ private static IEnumerable<T> ReadFromProtobufStream<T>(Stream stream, MessagePa
         }
 
         private static SonarQubeIssue ToSonarQubeIssue(ServerIssue issue) =>
-            new SonarQubeIssue(issue.Path, issue.Checksum, issue.Line, issue.Msg, issue.ModuleKey, issue.RuleKey,
+            new SonarQubeIssue(FilePathNormalizer.NormalizeSonarQubePath(issue.Path), issue.Checksum, issue.Line, issue.Msg, issue.ModuleKey, issue.RuleKey,
```
rita-g-sonarsource, line 91: We don't have any tests for it, and it's an old SQ version. Didn't know if I should add any or if it's obsolete.

SonarQube.Client/Api/V8_1/GetHotspotRequest.cs
```diff
@@ -46,7 +47,7 @@ protected override SonarQubeHotspot ParseResponse(string response)
                 serverResponse.Project.Key,
                 serverResponse.Project.Name,
                 serverResponse.Component.Key,
-                serverResponse.Component.Path,
+                FilePathNormalizer.NormalizeSonarQubePath(serverResponse.Component.Path),
```
duncanp-sonar, line 50: This doesn't feel right; we shouldn't be using the same property name (`ComponentPath`) as the server API now we're transforming the data. We should rename the property to something else e.g. 'RelativeFilePath', `ServerFilePath`, with an appropriate doc comment on the property.

It's not a problem with SonarQubeIssue class since the property is already called `FilePath`.

SonarQube.Client/Api/V5_10/GetIssuesRequest.cs
duncanp-sonar: I had the same problem with this class recently. I don't think it's worth the cost; it's an obsolete version of SQ, it's a minor change, and testing the protobuf API is painful.

SonarQube.Client/Api/V7_20/GetIssuesRequest.cs
```diff
@@ -79,7 +80,7 @@ protected override SonarQubeIssue[] ParseResponse(string response)
         }
 
         private SonarQubeIssue ToSonarQubeIssue(ServerIssue issue) =>
-            new SonarQubeIssue(ComputePath(issue.Component), issue.Hash, issue.Line, issue.Message, ComputeModuleKey(issue),
+            new SonarQubeIssue(FilePathNormalizer.NormalizeSonarQubePath(ComputePath(issue.Component)), issue.Hash, issue.Line, issue.Message, ComputeModuleKey(issue),
```
duncanp-sonar, line 83: I might have put the call to 'Normalize...' inside 'ComputePath' since it's part of the path processing, but it doesn't really make any difference.

SonarQube.Client/Models/SonarQubeIssue.cs
duncanp-sonar: I think it's worth a doc comment that the data is in Windows file format.

duncanp-sonar (approved):
Looks ok, except I think the property name `SonarQubeHotspot.ComponentPath` has to change since we're transforming the data.

I was expecting that some of the property names would change as we implemented the Open in IDE feature; it wasn't clear originally which properties we'd need, so I just exposed all of the server data unchanged with the same property names to make it clear what the mapping was.

rita-g-sonarsource:
> I was expecting that some of the property names would change as we implemented the Open in IDE feature; it wasn't clear originally which properties we'd need, so I just exposed all of the server data unchanged with the same property names to make it clear what the mapping was.

Any other properties you think we should change?

duncanp-sonar:
> Any other properties you think we should change?

Not so far, but I haven't really looked. There might be others we decide the data isn't in the form we want that we will have to transform and rename, and there might be some don't end up using and decide to drop. I think we should just look at it on a case by case basis at the point we start consuming the data.

PR #66, opened 2020-11-26
Dp/line hash

duncanp-sonar:
Update GetHotspot request to use public API:
* updated minimum SQ version to v8.6
* added extra data fields - line hash and rule description fields.

Relates to [SLVS #1905](https://github.com/SonarSource/sonarlint-visualstudio/issues/1905)

### DIFF
--- a/SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
+++ b/SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
@@ -35,7 +35,7 @@ namespace SonarQube.Client.Tests
         public async Task GetHotspot_Response_From_SonarQube()
         {
             const string hotspotKey = "AW9mgJw6eFC3pGl94Wrf";
-            await ConnectToSonarQube("8.1.0.0");
+            await ConnectToSonarQube("8.6.0.0");
 
             SetupRequest($"api/hotspots/show?hotspot={hotspotKey}", @"{
   ""key"": ""AW9mgJw6eFC3pGl94Wrf"",
--- a/SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
+++ b/SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
@@ -70,7 +70,6 @@ namespace SonarQube.Client.Tests
     ""endOffset"": 41
   },
   ""status"": ""TO_REVIEW"",
-  ""line"": 32,
   ""hash"": ""106b6a42b6099a78bbb233937f36c4bc"",
   ""message"": ""message"",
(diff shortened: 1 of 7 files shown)


SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
```diff
@@ -58,7 +58,10 @@ public async Task GetHotspot_Response_From_SonarQube()
     ""key"": ""java:S4787"",
     ""name"": ""rule-name"",
     ""securityCategory"": ""others"",
-    ""vulnerabilityProbability"": ""LOW""
+    ""vulnerabilityProbability"": ""LOW"",
+    ""riskDescription"": ""risk desc"",
+    ""vulnerabilityDescription"": ""vuln desc"",
+    ""fixRecommendations"": ""fix rec""
```
duncanp-sonar, line 64: Helpfully, the API also returns the three rule help fields, so I've included them too.

SonarQube.Client/Api/V8_6/GetHotspotRequest.cs
```diff
@@ -38,20 +36,29 @@ public class GetHotspotRequest : RequestBase<SonarQubeHotspot>, IGetHotspotReque
...
+
+            var rule = new SonarQubeHotspotRule(
+                serverResponse.Rule.Key,
+                serverResponse.Rule.Name,
+                serverResponse.Rule.SecurityCategory,
+                serverResponse.Rule.VulnerabilityProbability,
+                serverResponse.Rule.RiskDescription,
+                serverResponse.Rule.VulnerabilityDescription,
+                serverResponse.Rule.FixRecommendations
+            );
```
duncanp-sonar, line 48: I moved the rule definition data into a separate to make it easier to work with.

SonarQube.Client.Tests/SonarQubeService_GetHotspotAsync.cs
```diff
@@ -67,6 +70,8 @@ public async Task GetHotspot_Response_From_SonarQube()
     ""endOffset"": 41
   },
   ""status"": ""TO_REVIEW"",
+  ""line"": 32,
```
rita-g-sonarsource, line 73: Is the line field needed?
The test still asserts the line from textRange -- should we take it from here instead?

rita-g-sonarsource (approved):
Just the comment about the line number

SonarQube.Client/Api/V8_6/GetHotspotRequest.cs
rita-g-sonarsource: Good call, it started to be a lot

PR #68, opened 2020-12-08
Fetch taint vulnerabilities

rita-g-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/1990

### DIFF
--- a/SonarQube.Client.Tests/SonarQubeService_GetTaintVulnerabilitiesRequest.cs
+++ b/SonarQube.Client.Tests/SonarQubeService_GetTaintVulnerabilitiesRequest.cs
@@ -1,0 +1,648 @@
+﻿/*
+ * SonarQube Client
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarQube.Client.Models;
+
+namespace SonarQube.Client.Tests
+{
+    [TestClass]
+    public class SonarQubeService_GetTaintVulnerabilitiesRequest : SonarQubeService_TestBase
+    {
+        [TestMethod]
+        public void GetTaintVulnerabilities_NotConnected()
+        {
+            // No calls to Connect
+            // No need to setup request, the operation should fail
+            Func<Task<IList<SonarQubeIssue>>> func = async () =>
+                await service.GetTaintVulnerabilitiesAsync(It.IsAny<string>(), CancellationToken.None);
+
+            func.Should().ThrowExactly<InvalidOperationException>().And
+                .Message.Should().Be("This operation expects the service to be connected.");
+
+            logger.ErrorMessages.Should().Contain("The service is expected to be connected.");
+        }
+
+        [TestMethod]
+        public async Task GetTaintVulnerabilities_Response_From_SonarQube()
+        {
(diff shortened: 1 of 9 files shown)


SonarQube.Client/Api/V7_20/GetIssuesRequest.cs
```diff
@@ -80,11 +80,11 @@ protected override SonarQubeIssue[] ParseResponse(string response)
         }
 
         private SonarQubeIssue ToSonarQubeIssue(ServerIssue issue) =>
-            new SonarQubeIssue(FilePathNormalizer.NormalizeSonarQubePath(ComputePath(issue.Component)), issue.Hash, issue.Line, issue.Message, ComputeModuleKey(issue),
+            new SonarQubeIssue(ComputePath(issue.Component), issue.Hash, issue.Line, issue.Message, ComputeModuleKey(issue),
                 GetRuleKey(issue.CompositeRuleKey), issue.Status == "RESOLVED", ToIssueFlows(issue.Flows));
 
         private string ComputePath(string component) =>
-            componentKeyPathLookup[component].FirstOrDefault() ?? string.Empty;
+            FilePathNormalizer.NormalizeSonarQubePath(componentKeyPathLookup[component].FirstOrDefault() ?? string.Empty);
```
rita-g-sonarsource, line 87: I missed this when I added FilePathNormalizer but flows were never used up until now so no harm was done.

SonarQube.Client.Tests/SonarQubeService_GetTaintVulnerabilitiesRequest.cs
```diff
@@ -0,0 +1,648 @@
...
+            // should return only 2 taint out of 4 vulnerabilities
+            result.Count.Should().Be(2);
+
+            var taint1 = result[0];
+
+            taint1.RuleId.Should().Be("javasecurity:S2076");
+            taint1.Message.Should().Be("Refactor this code to not construct the OS command from tainted, user-controlled data.");
+            taint1.FilePath.Should().Be("src\\main\\java\\foo\\security\\injection\\CommandInjectionVulnerability.java");
+            taint1.Hash.Should().Be("1703916771e6abb765843e62a76fcb5a");
+            taint1.Line.Should().Be(17);
```
rita-g-sonarsource, line 619: FYI the API also returns TextRange field but we don't currently store it. The sink is in the locations as well so it's probably not that critical.

SonarQube.Client/Api/DefaultConfiguration.cs
```diff
@@ -50,7 +50,8 @@ public static RequestFactory Configure(RequestFactory requestFactory)
                 .RegisterRequest<IGetRoslynExportProfileRequest, V6_60.GetRoslynExportProfileRequest>("6.6")
                 .RegisterRequest<IGetOrganizationsRequest, V7_00.GetOrganizationsRequest>("7.0")
                 .RegisterRequest<IGetIssuesRequest, V7_20.GetIssuesRequestWrapper>("7.2")
-                .RegisterRequest<IGetHotspotRequest, V8_6.GetHotspotRequest>("8.6");
+                .RegisterRequest<IGetHotspotRequest, V8_6.GetHotspotRequest>("8.6")
+                .RegisterRequest<IGetTaintVulnerabilitiesRequest, V7_20.GetTaintVulnerabilitiesRequest>("7.2");
```
duncanp-sonar, line 54: Won't this mean that the "view taint vulnerabilities" feature will work against older versions of SonarQube from v7.2 onwards, and if so is it what we want?

From a technical point of view it probably doesn't make any difference, but it makes this more complicated from a marketing and support point of view. Probably worth checking what the PMs view is.

SonarQube.Client/Api/V7_20/GetTaintVulnerabilitiesRequest.cs
```diff
@@ -0,0 +1,62 @@
...
+            getIssuesRequest.Logger = Logger;
+            getIssuesRequest.ProjectKey = ProjectKey;
+            getIssuesRequest.Statuses = "OPEN,CONFIRMED,REOPENED";
+            getIssuesRequest.Types = "VULNERABILITY";
+
+            var vulnerabilities = await getIssuesRequest.InvokeAsync(httpClient, token);
+            WarnForApiLimit(vulnerabilities, getIssuesRequest);
+
+            var taintVulnerabilities = vulnerabilities
+                .Where(x => x.RuleId.Contains("security"))
```
duncanp-sonar, line 48: Wasn't there a `repo` field that contained just contained the repo key? That would make it clearer what the intent of the code was.

duncanp-sonar (approved):

SonarQube.Client/Api/DefaultConfiguration.cs
rita-g-sonarsource: Good point, I asked Marco. For now he suggests 7.2, he will let us know if this changes.

SonarQube.Client/Api/V7_20/GetTaintVulnerabilitiesRequest.cs
rita-g-sonarsource: No I don't see it in the response, and neither in the documentation: https://next.sonarqube.com/sonarqube/web_api/api/issues?query=issues#

PR #69, opened 2020-12-10
Add Severity and TextRange fields to SonarQubeIssue

rita-g-sonarsource:
Part of https://github.com/SonarSource/sonarlint-visualstudio/issues/1988

### DIFF
--- a/SonarQube.Client/Api/V5_10/GetIssuesRequest.cs
+++ b/SonarQube.Client/Api/V5_10/GetIssuesRequest.cs
@@ -88,7 +88,10 @@ namespace SonarQube.Client.Api.V5_10
         }
 
         private static SonarQubeIssue ToSonarQubeIssue(ServerIssue issue) =>
-            new SonarQubeIssue(FilePathNormalizer.NormalizeSonarQubePath(issue.Path), issue.Checksum, issue.Line, issue.Msg, issue.ModuleKey, issue.RuleKey,
-                issue.Status == "RESOLVED", flows: null);
+            new SonarQubeIssue(FilePathNormalizer.NormalizeSonarQubePath(issue.Path), issue.Checksum, issue.Msg,
+                issue.ModuleKey, issue.RuleKey, issue.Status == "RESOLVED",
+                severity: SonarQubeIssueSeverity.Unknown,
+                textRange: new IssueTextRange(issue.Line, issue.Line, 0, 0),
+                flows: null);
     }
 }
--- a/SonarQube.Client/Api/V7_20/GetIssuesRequest.cs
+++ b/SonarQube.Client/Api/V7_20/GetIssuesRequest.cs
@@ -80,8 +80,11 @@ namespace SonarQube.Client.Api.V7_20
         }
 
         private SonarQubeIssue ToSonarQubeIssue(ServerIssue issue) =>
-            new SonarQubeIssue(ComputePath(issue.Component), issue.Hash, issue.Line, issue.Message, ComputeModuleKey(issue),
-                GetRuleKey(issue.CompositeRuleKey), issue.Status == "RESOLVED", ToIssueFlows(issue.Flows));
+            new SonarQubeIssue(ComputePath(issue.Component), issue.Hash, issue.Message, ComputeModuleKey(issue),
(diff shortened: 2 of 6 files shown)


SonarQube.Client/Api/V5_10/GetIssuesRequest.cs
```diff
@@ -88,7 +88,10 @@ private static IEnumerable<T> ReadFromProtobufStream<T>(Stream stream, MessagePa
         }
 
         private static SonarQubeIssue ToSonarQubeIssue(ServerIssue issue) =>
-            new SonarQubeIssue(FilePathNormalizer.NormalizeSonarQubePath(issue.Path), issue.Checksum, issue.Line, issue.Msg, issue.ModuleKey, issue.RuleKey,
-                issue.Status == "RESOLVED", flows: null);
+            new SonarQubeIssue(FilePathNormalizer.NormalizeSonarQubePath(issue.Path), issue.Checksum, issue.Msg,
+                issue.ModuleKey, issue.RuleKey, issue.Status == "RESOLVED",
+                severity: SonarQubeIssueSeverity.Unknown,
```
rita-g-sonarsource, line 93: Technically the `severity` field is available, but this is such an old SQ version that I didn't bother (as this field is needed for taint and that will be for 7.2+ anyway).

SonarQube.Client/Api/V7_20/GetIssuesRequest.cs
```diff
@@ -102,8 +105,15 @@ protected override SonarQubeIssue[] ParseResponse(string response)
         private IssueLocation ToIssueLocation(ServerIssueLocation serverIssueLocation) =>
             new IssueLocation(ComputePath(serverIssueLocation.Component), serverIssueLocation.Component, ToIssueTextRange(serverIssueLocation.TextRange), serverIssueLocation.Message);
 
-        private static IssueTextRange ToIssueTextRange(ServerIssueTextRange serverIssueTextRange) =>
-            new IssueTextRange(serverIssueTextRange.StartLine, serverIssueTextRange.EndLine, serverIssueTextRange.StartOffset, serverIssueTextRange.EndOffset);
+        private static IssueTextRange ToIssueTextRange(ServerIssueTextRange serverIssueTextRange)
+        {
+            return serverIssueTextRange == null
```
rita-g-sonarsource, line 110: I checked and it seems that textRange field is available in 7.2+.
It indeed returns null when the issue is at file-level, same as the way Line was null.
E.g.:
https://peach.sonarsource.com/project/issues?id=akka-msbuild14&resolved=false&rules=csharpsquid%3AS4026

SonarQube.Client/Api/V7_20/GetIssuesRequest.cs
```diff
@@ -113,18 +123,28 @@ private class ServerIssue
...
+
             [JsonProperty("component")]
             public string Component { get; set; }
+
             [JsonProperty("subProject")]
             public string SubProject { get; set; }
+
             [JsonProperty("hash")]
             public string Hash { get; set; }
-            [JsonProperty("line")]
```
rita-g-sonarsource, line 122: Replaced with TextRange

SonarQube.Client/Models/SonarQubeIssue.cs
```diff
@@ -48,11 +49,12 @@ public class SonarQubeIssue
...
         public string FilePath { get; }
         public string Hash { get; }
-        public int? Line { get; }
         public string Message { get; }
         public string ModuleKey { get; }
         public string RuleId { get; }
         public bool IsResolved { get; }
+        public SonarQubeIssueSeverity Severity { get; }
+        public IssueTextRange TextRange { get; }
         public IReadOnlyList<IssueFlow> Flows { get; }
```
rita-g-sonarsource, line 58: Technically there is also the Type field, but since it's needed only for Taint for now, we already know in SLVS that it's "vulnerability".

duncanp-sonar:
I was going to ask whether we want to merge these changes into `master` directly or into a feature branch, but I see the last commit went directly into `master` so it would complicate things to start another branch now.
In general, I thinks it would be better to follow the same branch strategy during sprints as we use for the main repo, but we can decide what we want to do next sprint.

duncanp-sonar (approved):

PR #73, opened 2020-12-16
Added GetViewIssueUrl API

### LINKED ISSUE #72: [Taint] Add API to return a link to display an issue
Required by https://github.com/SonarSource/sonarlint-visualstudio/issues/1976

duncanp-sonar:
Fixes #72


SonarQube.Client/SonarQubeService.cs
```diff
@@ -329,6 +329,19 @@ public async Task<SonarQubeQualityProfile> GetQualityProfileAsync(string project
...
                 }, token);
 
+        public Uri GetViewIssueUrl(string projectKey, string issueKey)
+        {
+            EnsureIsConnected();
+
+            // The URL should be in the same form as the permalink generated by SQ/SC e.g
+            // SonarQube : http://localhost:9000/project/issues?id=security1&issues=AXZRhxr-9W_phHQ8Bzgn&open=AXZRhxr-9W_phHQ8Bzgn
+            // SonarCloud: https://sonarcloud.io/project/issues?id=sonarlint-visualstudio&issues=AW-EuNQbXwT7-YPcXpll&open=AW-EuNQbXwT7-YPcXpll
+            // Versioning: so far the format of the URL is the same across all versions from at least v6.7
```
duncanp-sonar, line 339: I tested on v6.7; I didn't test earlier versions because they have been out of support for years so I didn't think it was worth the effort (I didn't want to spend the time rolling back to Java 2 to run earlier versions).

SonarQube.Client/ISonarQubeService.cs
```diff
@@ -81,5 +81,11 @@ public interface ISonarQubeService
         Task<SonarQubeHotspot> GetHotspotAsync(string hotspotKey, CancellationToken token);
 
         Task<IList<SonarQubeIssue>> GetTaintVulnerabilitiesAsync(string projectKey, CancellationToken token);
+
+        /// <summary>
+        /// Returns the URI to view the specified issue on the server
+        /// </summary>
+        /// <remarks>The method does not check whether the project or issue exists or not</remarks>
+        Uri GetViewIssueUrl(string projectKey, string issueKey);
```
duncanp-sonar, line 89: Unfortunately this won't work for hotspots; there's a separate URL for them.
I was going to add an API for hotspots at the same time but decided to do it separately, partly because I couldn't decide how to handle requests for earlier versions of SQ i.e. whether to do any version checks or not. Perhaps there's no need; if the user is on an earlier version of SQ, there's no way in the UI they'd get to the point where we'd want to get a "view hotspot" URL.

rita-g-sonarsource (approved):

PR #76, opened 2020-12-17
Added IsSonarCloud property

### LINKED ISSUE #75: Add IsSonarCloud property to the service
### Description

An `IsSonarCloud` property would be useful for:
1) callers who e.g. want to display appropriate UI strings for the type of server being connected to, and 
2) the service itself, so it can correctly build the APIs strings if they vary between SonarQube and SonarCloud

duncanp-sonar:
Fixes #75

### DIFF
--- a/SonarQube.Client.Tests/Models/ConnectionInformationTests.cs
+++ b/SonarQube.Client.Tests/Models/ConnectionInformationTests.cs
@@ -84,9 +84,10 @@ namespace SonarQube.Client.Tests.Models
             var testSubject = new ConnectionInformation(new Uri(serverUrl), userName, securePwd);
             testSubject.Organization = org;
 
-            var cloneConnection = testSubject.Clone();
+            var cloneObj = (ICloneable)testSubject.Clone();
+            cloneObj.Should().BeOfType<ConnectionInformation>();
 
-            CheckPropertiesMatch(testSubject, cloneConnection);
+            CheckPropertiesMatch(testSubject, (ConnectionInformation)cloneObj);
         }
 
         [TestMethod]
--- a/SonarQube.Client.Tests/Models/ConnectionInformationTests.cs
+++ b/SonarQube.Client.Tests/Models/ConnectionInformationTests.cs
@@ -1,0 +1,62 @@
+﻿/*
+ * SonarQube Client
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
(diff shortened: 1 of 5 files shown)


SonarQube.Client/Models/ConnectionInformation.cs
```diff
@@ -38,49 +39,40 @@ public ConnectionInformation(Uri serverUri, string userName, SecureString passwo
...
 
-            this.ServerUri = FixSonarCloudUri(serverUri).EnsureTrailingSlash();
-            this.UserName = userName;
-            this.Password = password?.CopyAsReadOnly();
-            this.Authentication = AuthenticationType.Basic; // Only one supported at this point
+            ServerUri = FixSonarCloudUri(serverUri).EnsureTrailingSlash();
+            UserName = userName;
+            Password = password?.CopyAsReadOnly();
+            Authentication = AuthenticationType.Basic; // Only one supported at this point
+            IsSonarCloud = ServerUri == FixedSonarCloudUri;
```
duncanp-sonar, line 46: Added one new line. Preceding changes are removing `this.`.

SonarQube.Client.Tests/Models/ConnectionInformationTests.cs
```diff
@@ -0,0 +1,62 @@
...
+    [TestClass]
+    public class ConnectionInformationTests
+    {
+        [TestMethod]
+        [DataRow("http://localhost", "http://localhost/")]
+        [DataRow("http://localhost/", "http://localhost/")]
+        [DataRow("http://localhost:9000", "http://localhost:9000/")]
+        [DataRow("https://localhost:9000/", "https://localhost:9000/")]
+        [DataRow("https://local.sonarcloud.io", "https://local.sonarcloud.io/")]
+        public void Ctor_SonarQubeUrl_IsProcessedCorrectly(string inputUrl, string expectedUrl)
```
rita-g-sonarsource, line 37: Perhaps also `https://sonarcloud.io.local` ?

SonarQube.Client.Tests/SonarQubeService_Lifecycle.cs
```diff
@@ -51,24 +51,47 @@ public async Task Connect_To_SonarQube_Valid_Credentials()
...
 
-            action.Should().ThrowExactly<InvalidOperationException>()
-                .And.Message.Should().Be("Invalid credentials");
+            service.IsSonarCloud.Should().BeFalse();
+        }
 
-            service.IsConnected.Should().BeFalse();
-            service.SonarQubeVersion.Should().BeNull();
+        [TestMethod]
+        [DataRow("http://sonarcloud.io")]
```
rita-g-sonarsource, line 74: The test cases were already tested in the lower class.
The duplication is problematic.
I think it's better to pick just the obvious SQ/SC urls here and let the underlying class be tested for all the variations.

PR #78, opened 2020-12-17
Added required fields to SonarQubeIssue

### LINKED ISSUE #71: [Taint] Modify GetTaintIssues to return creation date
### Description

That taint MMF calls for issues to be displayed by creation date in descending order.
`api/issues/search` returns the `creationDate` but we don't currently pass it on.

Required fields:
* key, creationDate

TODO:
* decide whether to introduce a new taint-specific abstraction, since `creationDate` isn't required anywhere else (?probably not worth it?)
* check whether there are any other fields we could return that are either required or useful e.g. having `updateDate` might be useful so the user could see whether the issues has ever been looked at.
* check the date time format used by SonarQube (?vague memory that we had to do some additional formatting when handling date-times in telemetry)

duncanp-sonar:
Fixes #71

### DIFF
--- a/SonarQube.Client.Tests/SonarQubeService_GetTaintVulnerabilitiesRequest.cs
+++ b/SonarQube.Client.Tests/SonarQubeService_GetTaintVulnerabilitiesRequest.cs
@@ -492,8 +492,8 @@ namespace SonarQube.Client.Tests
 				""assign""
 			],
 			""comments"": [],
-			""creationDate"": ""2019-09-11T12:25:08+0000"",
-			""updateDate"": ""2019-09-16T10:35:19+0000"",
+			""creationDate"": ""2019-09-11T12:25:08+0100"",
+			""updateDate"": ""2019-09-16T10:35:19+1300"",
 			""type"": ""VULNERABILITY"",
 			""organization"": ""default-organization"",
 			""scope"": ""MAIN""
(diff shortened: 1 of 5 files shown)


SonarQube.Client.Tests/SonarQubeService_GetTaintVulnerabilitiesRequest.cs
```diff
@@ -492,8 +492,8 @@ public async Task GetTaintVulnerabilities_Response_From_SonarQube()
 				""assign""
 			],
 			""comments"": [],
-			""creationDate"": ""2019-09-11T12:25:08+0000"",
-			""updateDate"": ""2019-09-16T10:35:19+0000"",
+			""creationDate"": ""2019-09-11T12:25:08+0100"",
+			""updateDate"": ""2019-09-16T10:35:19+1300"",
```
duncanp-sonar, line 496: I modified these values to include time zone information to test that the time zone is taken in to account when deserializing.

SonarQube.Client/Api/V5_10/GetIssuesRequest.cs
```diff
@@ -88,9 +88,11 @@ private static IEnumerable<T> ReadFromProtobufStream<T>(Stream stream, MessagePa
         }
 
         private static SonarQubeIssue ToSonarQubeIssue(ServerIssue issue) =>
-            new SonarQubeIssue(FilePathNormalizer.NormalizeSonarQubePath(issue.Path), issue.Checksum, issue.Msg,
+            new SonarQubeIssue(issue.Key, FilePathNormalizer.NormalizeSonarQubePath(issue.Path), issue.Checksum, issue.Msg,
```
duncanp-sonar, line 91: The change in this line is causing the "low" code coverage (one uncovered line out of six). There aren't any tests for this legacy method, and I'm not intending to add any as part of this change.

SonarQube.Client/Api/V7_20/GetIssuesRequest.cs
```diff
@@ -145,6 +150,12 @@ private class ServerIssue
             [JsonProperty("textRange")]
             public ServerIssueTextRange TextRange { get; set; }
 
+            [JsonProperty("creationDate")]
+            public virtual DateTimeOffset CreationDate { get; set; }
```
duncanp-sonar, line 154: The web service returns data in a format that is handled correctly by the JSON deserializer without us having to do anything special.

This library does have a special `JavaDateConverter` class. However, it looks like this exists to work round a bug when sending data to `GetNotificationsRequest`. The date parameter sent to that API is encoded in a special way by the `JavaDateConverter`. There is no special handling of the date values returned by that API.

duncanp-sonar:
Added the `key`, `creationDate` and `lastUpdate` fields to `SonarQubeIssue`.

There are a couple of other fields that I think would be useful to the user when looking at taint issues, namely `assignee`, `author` and `comments`. However, we can add them later if necessary.

rita-g-sonarsource (approved):

Issue #80, opened 2021-01-05
Support separate selection of SonarCloud APIs


duncanp-sonar:
### Description

Historically SonarQube and SonarCloud supported the same APIs and the same versioning scheme, so there was no need for the library to differentiate between SonarQube/SonarCloud.

This is no longer the case. The server version numbering of SonarCloud has already diverged from that of SonarQube so the version number alone can no longer be used as the sole way to select the appropriate API call to make.

The library separate selection logic for SonarCloud so it can pick a completely different set of APIs if necessary.

PR #81, opened 2021-01-14
Added abstractions to support handling SonarCloud requests separately

duncanp-sonar:
Added IRequestHandler interface.
Combined the existing `SonarQubeVersion` and `IsSonarCloud` properties into a new `ServerInfo` class.
Updated the request handler to use `ServerInfo` rather than `Version`.

Relates to #80 / https://github.com/SonarSource/sonarlint-visualstudio/issues/2060

### DIFF
--- a/SonarQube.Client/Requests/IRequestFactory.cs
+++ b/SonarQube.Client/Requests/IRequestFactory.cs
@@ -1,0 +1,30 @@
+﻿/*
+ * SonarQube Client
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+namespace SonarQube.Client.Requests
+{
+    /// <summary>
+    /// Factory that creates request instance appropriate to the supplied server type/version
+    /// </summary>
+    internal interface IRequestFactory
+    {
+        TRequest Create<TRequest>(ServerInfo serverInfo) where TRequest : IRequest;
+    }
+}
--- a/SonarQube.Client/Requests/RequestFactory.cs
+++ b/SonarQube.Client/Requests/RequestFactory.cs
@@ -25,7 +25,12 @@ using SonarQube.Client.Logging;
 
 namespace SonarQube.Client.Requests
 {
-    public class RequestFactory
+    internal interface IRequestFactory
+    {
+        TRequest Create<TRequest>(Version version) where TRequest : IRequest;
+    }
+
+    internal class RequestFactory : IRequestFactory
     {
         /// <summary>
         /// Map between request type and a list with versioned request implementation factories.
--- a/SonarQube.Client/Requests/RequestFactory.cs
+++ b/SonarQube.Client/Requests/RequestFactory.cs
@@ -25,11 +25,6 @@ using SonarQube.Client.Logging;
 
 namespace SonarQube.Client.Requests
 {
-    internal interface IRequestFactory
-    {
-        TRequest Create<TRequest>(Version version) where TRequest : IRequest;
-    }
-
     internal class RequestFactory : IRequestFactory
     {
         /// <summary>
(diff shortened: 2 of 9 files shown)


duncanp-sonar:
Currently the 'RequestFactory' is a concrete class that decides which API to call based entirely on the server version.
This is no longer flexible enough; SonarQube and SonarCloud versioning have already diverged, and the actual web APIs may diverge in the future.

The `RequestFactory` needs to be an abstraction so we can create different factories for SonarQube and SonarCloud.
It also needs to use more information than just the version to decide which API to call.

This PR adds some of the necessary abstractions so the `SonarQubeService` is no longer directly dependent on the existing concrete `RequestFactory`, and it also supplies more information to the factory to allow it to decide which API to call.

SonarQube.Client/Requests/IRequestFactory.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace SonarQube.Client.Requests
+{
+    /// <summary>
+    /// Factory that creates request instance appropriate to the supplied server type/version
+    /// </summary>
+    internal interface IRequestFactory
+    {
+        TRequest Create<TRequest>(ServerInfo serverInfo) where TRequest : IRequest;
+    }
```
duncanp-sonar, line 29: New abstraction to decouple the `ISonarQubeService`  from the single existing factory. The other important change is that the supplied parameter is now `ServerInfo` rather than just `Version`.

Logically, `ServerInfo` should contain all of the info a factory should need to decide to which API to call. In practice this means the version and whether the server is SQ or SC.

SonarQube.Client/Requests/RequestFactory.cs
```diff
@@ -25,7 +25,7 @@
 
 namespace SonarQube.Client.Requests
 {
-    public class RequestFactory
+    internal class RequestFactory : IRequestFactory
```
duncanp-sonar, line 28: Miscellaneous: didn't need to be public

SonarQube.Client/Requests/RequestFactory.cs
```diff
@@ -94,17 +94,18 @@ public RequestFactory(ILogger logger)
         /// Creates a new TRequest implementation for the specified SonarQube version.
         /// </summary>
         /// <typeparam name="TRequest">The type of the request implementation to create.</typeparam>
-        /// <param name="version">
+        /// <param name="serverInfo">
         /// SonarQube version to return a request implementation for. When the provided value is null
```
duncanp-sonar, line 98: ```/// SonarQube version...``` comment needs to be updated.

SonarQube.Client/Requests/RequestFactory.cs
```diff
@@ -94,17 +94,18 @@ public RequestFactory(ILogger logger)
...
         /// </param>
         /// <returns>New the newest TRequest implementation for the specified SonarQube version.</returns>
-        public TRequest Create<TRequest>(Version version)
+        public TRequest Create<TRequest>(ServerInfo serverInfo)
             where TRequest : IRequest
         {
             SortedList<Version, Func<IRequest>> map;
             if (registrations.TryGetValue(typeof(TRequest), out map))
             {
+                var version = serverInfo?.Version;
```
duncanp-sonar, line 108: It seems weird that we would be doing a lookup with a null version. However, this is a special chicken-and-egg case that arises when initially connecting - at that point, we don't know what the version is because we haven't asked the server yet...

SonarQube.Client/SonarQubeService.cs
```diff
@@ -58,14 +54,14 @@ public async Task<bool> HasOrganizations(CancellationToken token)
...
-        public bool IsSonarCloud { get; private set; }
+        public ServerInfo ServerInfo { get; private set; }
 
         public SonarQubeService(HttpMessageHandler messageHandler, string userAgent, ILogger logger)
             : this(messageHandler, DefaultConfiguration.Configure(new RequestFactory(logger)), userAgent, logger)
         {
         }
 
-        public SonarQubeService(HttpMessageHandler messageHandler, RequestFactory requestFactory, string userAgent, ILogger logger)
+        internal /* for testing */ SonarQubeService(HttpMessageHandler messageHandler, IRequestFactory requestFactory, string userAgent, ILogger logger)
```
duncanp-sonar, line 64: Didn't need to be public

rita-g-sonarsource:
> Currently the 'RequestFactory' is a concrete class that decides which API to call based entirely on the server version.
> This is no longer flexible enough; SonarQube and SonarCloud versioning have already diverged, and the actual web APIs may diverge in the future.
> 
> The `RequestFactory` needs to be an abstraction so we can create different factories for SonarQube and SonarCloud.
> It also needs to use more information than just the version to decide which API to call.
> 
> This PR adds some of the necessary abstractions so the `SonarQubeService` is no longer directly dependent on the existing concrete `RequestFactory`, and it also supplies more information to the factory to allow it to decide which API to call.
> 
> Next steps:
> My intention is to add the following in subsequent PR(s):
> 
> * a new factory class that handles SonarCloud
> * an aggregating factory class that sits on top of the current "versioned" request and the new SonarCloud request factory.
> * rename existing classes where appropriate e.g. `RequestFactory` to something like `SonarQubeRequestFactory` or `VersionedRequestFactory`.

I can't quite put my finger on it, but something about ServerInfo and the usages feels odd. 
Since the SC factory is not versioned, it feels odd to have the version as part of the API. The version should be contained inside the SQ factory, and not exposed as part of the API, as it's not the case for all factories anymore.
You mentioned that you want to create a factory on top, how would that look like?

Ideally, I'd want the SonarQubeService to do something like
```
var requestFactory = requestFactoryProvider.Get(IsSonarCloud);
var request = requestFactory.Create<T>();
```
And in each factory, do something like:
```
class RequestFactoryProvider
{
     IRequestFactory Get(bool isSonarCloud)
     {
         if (isSonarCloud)
         {
               return new SonarCloudFactory();
         }
         var sonarQubeFactory = new SonarQubeFactory();
         sonarQubeFactory.Initialize(); // fetch the version
         return sonarQubeFactory;
     }
}
```

```
class SonarQubeFactory : IRequestFactory
{
     private Version version;
     
     void Initialize()
     {
          version = GetVersion();
     }
     
      IRequest Create<T>()
      {
          // use the version
      }
}
```

Ignore the exact implementation, this is a general concept. My point is that I'd like the version to be contained inside the SQ factory. It would make for a cleaner API. WDYT?

duncanp-sonar:
Yes; I think there is a missing abstraction to handle the initial connection, and I agree that `IRequestHandler.Create<T>` call shouldn't  have any parameters. However, I think moving to that design can be done in stages, and it doesn't all have to be done now. The key change that is required at this point is the ability to plug in a different request factory for SonarCloud. That can be done easily at this point without changing the `Create` method signature.

rita-g-sonarsource (approved):

PR #82, opened 2021-01-15
Added aggregating request factory

duncanp-sonar:
Relates to #80

### DIFF
--- a/SonarQube.Client/Requests/AggregatingRequestFactory.cs
+++ b/SonarQube.Client/Requests/AggregatingRequestFactory.cs
@@ -20,15 +20,26 @@
 
 namespace SonarQube.Client.Requests
 {
-    /// <summary>
-    /// Factory that creates request instance appropriate to the supplied server type/version
-    /// </summary>
-    internal interface IRequestFactory
+    internal class AggregatingRequestFactory : IRequestFactory
     {
-        /// <summary>
-        /// Creates and returns a new instance of the specified request
-        /// </summary>
-        /// <remarks>If the factory does not handle the specified server type it should return null</remarks>
-        TRequest Create<TRequest>(ServerInfo serverInfo) where TRequest : class, IRequest;
+        private readonly IRequestFactory[] requestFactories;
+
+        public AggregatingRequestFactory(params IRequestFactory[] requestFactories)
+        {
+            this.requestFactories = requestFactories;
+        }
+
+        public TRequest Create<TRequest>(ServerInfo serverInfo) where TRequest : class, IRequest
+        {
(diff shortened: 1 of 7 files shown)


SonarQube.Client/Requests/AggregatingRequestFactory.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            this.requestFactories = requestFactories;
+        }
+
+        public TRequest Create<TRequest>(ServerInfo serverInfo) where TRequest : class, IRequest
+        {
+            foreach (var factory in requestFactories)
+            {
+                var newRequest = factory.Create<TRequest>(serverInfo);
+                if (newRequest != null)
```
duncanp-sonar, line 37: The main reason I added the `class` constraint was so this comparison is against `null` rather than `default(T)`.
We know the instances are all going to be classes, so it seemed clearer to make it explicit.

duncanp-sonar:
These changes are just enough to support multiple request factories.

Next steps: add a new factory to handle SonarCloud-specific requests. I may have to tweak the `SonarQubeService.ConnectAsync` logic slightly too.

SonarQube.Client/Requests/AggregatingRequestFactory.cs
```diff
@@ -0,0 +1,45 @@
...
+        private readonly IRequestFactory[] requestFactories;
+
+        public AggregatingRequestFactory(params IRequestFactory[] requestFactories)
+        {
+            this.requestFactories = requestFactories;
+        }
+
+        public TRequest Create<TRequest>(ServerInfo serverInfo) where TRequest : class, IRequest
+        {
+            foreach (var factory in requestFactories)
```
rita-g-sonarsource, line 34: I'm a bit confused why would we need this?
The server is always either SC or SQ.
This seems to attempt first call SQ, and only if it fails, call SC (or vice-versa). That doesn't seem right.
We should be calling whatever server is configured and only that one. If the API is not available in it, we should fail, not fallback to a different server API. It's also entirely possible that the API would be available in both versions but return different values.

I don't think we should be iterating over the factories.
We should have only 2, get one of them at the start of the connection, and then use only that one until the connection is reset.

duncanp-sonar: Yep, you're right. That's what I originally intended to do, but managed to confuse myself into thinking it wouldn't work without more refactoring. However, I've just been looking at `ConnectAsync` and it should be doable.

FYI in the aggregating approach the factory implementations would have checked the server type and returned null if they didn't handle the server type.

PR #84, opened 2021-01-15
Added unversioned request factory

duncanp-sonar:
Added unversioned request factory
Added configuration for SonarCloud

Relates to #80

### DIFF
--- a/SonarQube.Client.Tests/Requests/UnversionedRequestFactoryTests.cs
+++ b/SonarQube.Client.Tests/Requests/UnversionedRequestFactoryTests.cs
@@ -1,0 +1,86 @@
+﻿/*
+ * SonarQube Client
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarQube.Client.Logging;
+using SonarQube.Client.Requests;
+using SonarQube.Client.Tests.Infra;
+
+namespace SonarQube.Client.Tests.Requests
+{
+    [TestClass]
+    public class UnversionedRequestFactoryTests
+    {
+        [TestMethod]
+        public void Create_Throws_When_Not_Registered()
+        {
+            var logger = new TestLogger();
+            var factory = new UnversionedRequestFactory(logger);
+            Action action = () => factory.Create<ITestRequest>(null);
+            action.Should().ThrowExactly<InvalidOperationException>()
+                .WithMessage("Could not find factory for 'ITestRequest'.");
+
+            logger.ErrorMessages.Should().Contain(new[] { "Could not find factory for 'ITestRequest'." });
+        }
+
+        [TestMethod]
+        public void Create_Returns_New_Instance()
+        {
+            var logger = new TestLogger();
+            var factory = new UnversionedRequestFactory(logger);
+            factory.RegisterRequest<ITestRequest, TestRequest>();
+            factory.RegisterRequest<IAnotherRequest, AnotherRequest>();
+
+            factory.Create<ITestRequest>(null).Should().BeOfType<TestRequest>();
+            factory.Create<IAnotherRequest>(null).Should().BeOfType<AnotherRequest>();
+        }
+
+        [TestMethod]
+        public void Register_Same_RequestInterface_Throws()
+        {
+            var logger = new TestLogger();
+            var factory = new UnversionedRequestFactory(logger);
+            factory.RegisterRequest<ITestRequest, TestRequest>();
+
+            Action action = () => factory.RegisterRequest<ITestRequest, TestRequest>();
+
+            action.Should().ThrowExactly<InvalidOperationException>()
+                .WithMessage("Registration for ITestRequest already exists.");
+
+            logger.ErrorMessages.Should().Contain(new[] { "Registration for ITestRequest already exists." });
+        }
+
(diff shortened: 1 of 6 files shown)


SonarQube.Client/Api/DefaultConfiguration.cs
```diff
@@ -24,7 +24,7 @@ namespace SonarQube.Client.Api
 {
     internal static class DefaultConfiguration
     {
-        public static RequestFactory Configure(RequestFactory requestFactory)
+        public static RequestFactory ConfigureSonarQube(RequestFactory requestFactory)
```
duncanp-sonar, line 27: Renamed to make it clear what is being configured.

SonarQube.Client/Api/DefaultConfiguration.cs
```diff
@@ -54,5 +54,29 @@ public static RequestFactory Configure(RequestFactory requestFactory)
                 .RegisterRequest<IGetTaintVulnerabilitiesRequest, V8_6.GetTaintVulnerabilitiesRequest>("8.6");
             return requestFactory;
         }
+
+        public static UnversionedRequestFactory ConfigureSonarCloud(UnversionedRequestFactory requestFactory)
```
duncanp-sonar, line 58: Initially I was going to implement a specific `IRequestFactory` class for SonarCloud. It wouldn't have needed to support dynamic registration, and the mapping of requests-to-factory functions could just have been a `readonly` dictionary that class.

In the end I decided it would be useful to have a single file in which we could see what was registered for each service i.e. this one. I'm not sure it was the right decision - time will tell.

That decision meant I needed to add a new type of "unversioned factory". This is actually just a cut-down version of the current `RequestFactory` that doesn't support registering multiple versions for each request.

SonarQube.Client/Requests/IRequestFactorySelector.cs
```diff
@@ -42,7 +42,7 @@ public IRequestFactory Select(bool isSonarCloud, ILogger logger)
             }
 
             logger.Debug("Selected SonarQube request factory");
-            return DefaultConfiguration.Configure(new RequestFactory(logger));
+            return DefaultConfiguration.ConfigureSonarQube(new RequestFactory(logger));
```
duncanp-sonar, line 45: Note: this class doesn't yet return a SonarCloud-specific factory.
That change will break some of the existing tests, so I left it for another PR.

SonarQube.Client/Api/DefaultConfiguration.cs
```diff
@@ -54,5 +54,29 @@ public static RequestFactory Configure(RequestFactory requestFactory)
...
+                .RegisterRequest<IGetQualityProfileChangeLogRequest, V6_50.GetQualityProfileChangeLogRequest>()
+                .RegisterRequest<IGetQualityProfilesRequest, V6_50.GetQualityProfilesRequest>()
+                .RegisterRequest<IGetNotificationsRequest, V6_60.GetNotificationsRequest>()
+                .RegisterRequest<IGetRoslynExportProfileRequest, V6_60.GetRoslynExportProfileRequest>()
+                .RegisterRequest<IGetOrganizationsRequest, V7_00.GetOrganizationsRequest>()
+                .RegisterRequest<IGetIssuesRequest, V7_20.GetIssuesRequestWrapper>()
+                .RegisterRequest<IGetHotspotRequest, V8_6.GetHotspotRequest>()
+                .RegisterRequest<IGetTaintVulnerabilitiesRequest, V8_6.GetTaintVulnerabilitiesRequest>();
+
+            return requestFactory;
```
duncanp-sonar, line 79: We don't currently have any SonarCloud-specific APIs, so this list is just the newest version of each current request implemented for SonarQube.

I created it by copying the entire list from from the SonarQube registration and running the test. Each time it failed because of a duplicate resgistration I deleted the earliest duplicate and ran it again.

SonarQube.Client/Requests/UnversionedRequestFactory.cs
```diff
@@ -0,0 +1,84 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Collections.Generic;
+using SonarQube.Client.Logging;
+
+namespace SonarQube.Client.Requests
+{
+    internal class UnversionedRequestFactory : IRequestFactory
```
duncanp-sonar, line 27: This is a cut-and-paste of the existing request factory with the complicated bits cut out i.e. the support for multiple versioned request types.

SonarQube.Client.Tests/Requests/UnversionedRequestFactoryTests.cs
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+using System;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarQube.Client.Logging;
+using SonarQube.Client.Requests;
+using SonarQube.Client.Tests.Infra;
+
+namespace SonarQube.Client.Tests.Requests
```
duncanp-sonar, line 28: I copied and pasted the tests for the existing RequestFactory and cut out the bits that were no longer appropriate.

SonarQube.Client.Tests/Requests/UnversionedRequestFactoryTests.cs
```diff
@@ -0,0 +1,86 @@
...
+        [TestMethod]
+        public void Create_Returns_New_Instance()
+        {
+            var logger = new TestLogger();
+            var factory = new UnversionedRequestFactory(logger);
+            factory.RegisterRequest<ITestRequest, TestRequest>();
+            factory.RegisterRequest<IAnotherRequest, AnotherRequest>();
+
+            factory.Create<ITestRequest>(null).Should().BeOfType<TestRequest>();
+            factory.Create<IAnotherRequest>(null).Should().BeOfType<AnotherRequest>();
```
duncanp-sonar, line 54: I tweaked this test to add and request two different types. If you only add one type you can't tell if the test is returning the correct one or not.

SonarQube.Client/Requests/UnversionedRequestFactory.cs
```diff
@@ -0,0 +1,84 @@
...
+                    $"Registration for {typeof(TRequest).Name} already exists.");
+            }
+
+            requestToFactoryMap[typeof(TRequest)] = () => new TRequestImpl();
+            logger.Debug($"Registered {typeof(TRequestImpl).FullName}");
+
+            return this;
+        }
+
+        public TRequest Create<TRequest>(ServerInfo serverInfo) where TRequest : IRequest
```
rita-g-sonarsource, line 66: I think the ServerInfo class can go now: it's not needed at all for SC, and for SQ we can pass the version in the ctor - the version would be the same for every `Create` call, as we would create a new factory each time the connection changes.

rita-g-sonarsource (approved):

SonarQube.Client/Requests/UnversionedRequestFactory.cs
duncanp-sonar: As discussed offline, this would require changing the factory to make calls to the server. I'll add some tickets to the backlog.

PR #88, opened 2021-01-22
Added GetSourceCode API

### LINKED ISSUE #86: Add API to return source code for a file
### Description

Add an API to return the source code for a file.
Input(s): the project key, the relative file path
Output(s): the raw source code for the file

#### Server-side web APIs:
* SonarQube: `api/sources/raw?key={fileKey}` e.g. my_project:src/foo/Bar.php. Sinvce v 5.0.
* SonarCloud: [`api/sources/raw`](https://sonarcloud.io/web_api/api/sources/raw)

### Background
SLVS needs the hashes for issue secondary locations in order to determine whether the code in the IDE has changed since the issue was raised. However, neither SonarQube nor SonarCloud currently has an API to return secondary location hashes. This API is a workaround to allow SLVS to calculate the hashes on the client-side.

duncanp-sonar:
Fixed #86

### DIFF
--- a/SonarQube.Client.Tests/CallRealServerTestHarness.cs
+++ b/SonarQube.Client.Tests/CallRealServerTestHarness.cs
@@ -1,0 +1,100 @@
+﻿/*
+ * SonarQube Client
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Net.Http;
+using System.Security;
+using System.Threading;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarQube.Client.Models;
+using SonarQube.Client.Tests.Infra;
+
+// Dummy tests. These exist to make it easy to make real SonarQube/SonarCloud
+// calls during development
+//
+// Usage:
+// * uncomment the [TestClass] attribute
+// * if calling SonarCloud, set a valid SonarCloud token
+
+namespace SonarQube.Client.Tests
+{
+    //[TestClass]
+    public class CallRealServerTestHarness
+    {
+        [TestMethod]
+        public async Task Call_Real_SonarQube()
+        {
+            var url = new Uri("http://localhost:9000");
+
+            string userName = "admin";
+            var password = new SecureString();
+            password.AppendChar('a');
+            password.AppendChar('d');
+            password.AppendChar('m');
+            password.AppendChar('i');
+            password.AppendChar('n');
+
+            var connInfo = new ConnectionInformation(url, userName, password);
+
+            var service = new SonarQubeService(new HttpClientHandler(), "agent", new TestLogger());
+            try
+            {
+                await service.ConnectAsync(connInfo, CancellationToken.None);
+
+                // Example
+                string fileKey = "junk:MyClass.cs";
+                var result = await service.GetSourceCodeAsync(fileKey, CancellationToken.None);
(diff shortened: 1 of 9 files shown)


SonarQube.Client.Tests/CallRealServerTestHarness.cs
```diff
@@ -0,0 +1,100 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SonarQube.Client.Models;
+using SonarQube.Client.Tests.Infra;
+
+// Dummy tests. These exist to make it easy to make real SonarQube/SonarCloud
+// calls during development
+//
+// Usage:
+// * uncomment the [TestClass] attribute
+// * if calling SonarCloud, set a valid SonarCloud token
```
duncanp-sonar, line 36: Miscellaneous. Simple test harness to make it easier to test the new code against a live server.

SonarQube.Client.Tests/SonarQubeService_GetSourceCodeAsync.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            var result = await service.GetSourceCodeAsync("my_project:src/foo/Bar.php", CancellationToken.None);
+
+            messageHandler.VerifyAll();
+
+            result.Should().Be(sourceCode);
+        }
+
+        [TestMethod]
+        public async Task Get_NotFound()
```
duncanp-sonar, line 59: The next two tests are common to all requests.

SonarQube.Client/ISonarQubeService.cs
```diff
@@ -93,5 +93,11 @@ public interface ISonarQubeService
         /// </summary>
         /// <remarks>The method does not check whether the project or hotspot exists or not</remarks>
         Uri GetViewHotspotUrl(string projectKey, string hotspotKey);
+
+        /// <summary>
+        /// Returns the source code for the specified file
+        /// </summary>
+        /// <param name="fileKey">e.g. my_project:src/foo/Bar.php</param>
```
duncanp-sonar, line 100: The format of the `fileKey` property is a bit ugly. I debated whether to have the caller pass the project key and relative file paths separately, since that is what the parameter appears to be.
However, I decided to stick to with the single parameter to start with. If it's inconvenient to work with from the client side then I'll change it later.

SonarQube.Client/Api/V5_00/GetSourceCodeRequest.cs
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Newtonsoft.Json;
+using SonarQube.Client.Requests;
+
+namespace SonarQube.Client.Api.V5_00
```
duncanp-sonar, line 24: Fortunately, both SC and SQ have the same API currently.

SQ v5.0 is the earliest supported version. Even though the only consumer of this API for now will be by the taint feature which requires v8.x, I didn't see a reason to only make this API available in newer versions of SonarQube.

SonarQube.Client/Api/IGetSourceCode.cs
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using SonarQube.Client.Requests;
+
+namespace SonarQube.Client.Api
+{
+    public interface IGetSourceCode : IRequest<string>
```
duncanp-sonar, line 25: The web returns a simple string - no need to mess about with deserializing JSON.

rita-g-sonarsource:
@duncanp-sonar I'm sorry for the late review, missed the notification.

SonarQube.Client/ISonarQubeService.cs
rita-g-sonarsource: I don't think it should be on the caller's side, it's an internal implementation detail of the API. But ok if you want to see how it looks.

SonarQube.Client/Api/IGetSourceCode.cs
rita-g-sonarsource, line 25: I see that the naming convention for the other requests is IXRequest

SonarQube.Client.Tests/SonarQubeService_GetSourceCodeAsync.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        [TestMethod]
+        public async Task Get_V5_0_ExampleFromSonarQube()
+        {
+            await ConnectToSonarQube("5.0.0.0");
+
+            const string sourceCode = @"package org.sonar.check;
+
+public enum Priority {
+/**
```
rita-g-sonarsource, line 45: I was extremely confused by this comment and didn't understand why is there an enum declaration in the middle of the test.
Can we go with a simpler source code, one that doesn't have comments in it?

rita-g-sonarsource (approved):

Issue #89, opened 2021-01-25
Modify GetIssues to optionally return line hashes for secondary locations


duncanp-sonar:
### Description

The SonarQube/SonarCloud issue web APIs do not currently return the line hashes for secondary locations.
Ticket #86 adds an API to fetch the source code for a file so the hashes can be calculated client-side.

This ticket is to extend the GetIssues API so the caller can choose whether the secondary location hashes should be returned or not. By default, the secondary line hashes will *not* be returned because calculating the line hash is a potentially expensive operation: it will generate an extra web call for each unique secondary location file.

Depends on #86

PR #90, opened 2021-01-25
Moved checksum calculator from SLVS to client library

duncanp-sonar:
Required for #89

### DIFF
--- a/SonarQube.Client.Tests/ChecksumCalculatorTests.cs
+++ b/SonarQube.Client.Tests/ChecksumCalculatorTests.cs
@@ -1,0 +1,113 @@
+﻿/*
+ * SonarQube Client
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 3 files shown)


SonarQube.Client.Tests/ChecksumCalculatorTests.cs
```diff
@@ -0,0 +1,113 @@
...
+{
+    [TestClass]
+    public class ChecksumCalculatorTests
+    {
+        [TestMethod]
+        public void Checksum_ThrowsOnNull()
+        {
+            Action act = () => ChecksumCalculator.Calculate(null);
+
+            act.Should().ThrowExactly<ArgumentNullException>().And.ParamName.Should().Be("text");
```
duncanp-sonar, line 35: This test has changed slightly so it doesn't use the SLVS `Exceptions.Expect` helper class.

duncanp-sonar:
The checksum calculator will be needed to calculate the line hashes. It also makes sense logically for the logic to be in the client library. I'll remove it from SLVS when updating the submodule

No changes in the product code - just the namespaces and copyright notice.

rita-g-sonarsource (approved):

PR #91, opened 2021-01-25
Added API to return branch info

### LINKED ISSUE #87: Add API to return branch information
### Description

Add an API to return information about the analyzed branches.
Input(s): the project key
Output(s): a list of [branch name, last analyzed date, whether the branch is the main branch or not]

#### Server-side web APIs:
* SonarQube: `api/project_branches/list?project={projectKey}` Since v 6.6
* SonarCloud: [`api/project_branches/list`](https://sonarcloud.io/web_api/api/project_branches/list)

#### Sample response
```
{
  "branches": [
    {
      "name": "feature/foo",
      "isMain": false,
      "type": "BRANCH",
      "status": {
        "qualityGateStatus": "OK"
      },
      "analysisDate": "2017-04-03T13:37:00+0100",
      "excludedFromPurge": false
    },
    {
      "name": "master",
      "isMain": true,
      "type": "BRANCH",
      "status": {
        "qualityGateStatus": "ERROR"
      },
      "analysisDate": "2017-04-01T01:15:42+0100",
      "excludedFromPurge": true
    }
  ]
}
```

### Background
The initial requirement from SLVS is to be able to display the date when the main branch of a project was last analyzed. This API contains that information and more. It's likely that more branch information will be required in the future, so it makes sense to return the list of branches and let SLVS filter out the information it requires.

duncanp-sonar:
Fixes #87


SonarQube.Client.Tests/Requests/DefaultConfiguration_Configure.cs
```diff
@@ -94,6 +95,7 @@ public void ConfigureSonarCloud_Writes_Debug_Messages()
                     "Registered SonarQube.Client.Api.V6_50.GetQualityProfilesRequest",
                     "Registered SonarQube.Client.Api.V6_60.GetNotificationsRequest",
                     "Registered SonarQube.Client.Api.V6_60.GetRoslynExportProfileRequest",
+                    "Registered SonarQube.Client.Api.V6_60.GetProjectBranchesRequest",
```
duncanp-sonar, line 98: SonarQube and SonarCloud currently support the same API.

SonarQube.Client.Tests/SonarQubeService_GetProjectBranchesAsync.cs
```diff
@@ -0,0 +1,113 @@
...
+            result[0].IsMain.Should().Be(false);
+            result[0].LastAnalysisTimestamp.Should().Be(new DateTimeOffset(2017, 4, 3, 13, 37, 0, TimeSpan.FromHours(1)));
+
+            result[1].Name.Should().Be("master");
+            result[1].IsMain.Should().Be(true);
+            result[1].LastAnalysisTimestamp.Should().Be(new DateTimeOffset(2018, 12, 1, 1, 15, 42, TimeSpan.FromHours(-4)));
+        }
+
+        [TestMethod]
+        public async Task Get_NotFound()
```
duncanp-sonar, line 83: This test and the next one are the common tests for all requests.

SonarQube.Client.Tests/SonarQubeService_GetProjectBranchesAsync.cs
```diff
@@ -0,0 +1,113 @@
...
+      },
+      ""analysisDate"": ""2017-04-03T13:37:00+0100"",
+      ""excludedFromPurge"": false
+    },
+    {
+      ""name"": ""master"",
+      ""isMain"": true,
+      ""type"": ""BRANCH"",
+      ""status"": {
+        ""qualityGateStatus"": ""ERROR""
```
duncanp-sonar, line 58: The `type` and `status` fields might be useful in the future, but it isn't documented what values could be, and SonarCloud appears to return different values from SonarQube. So I didn't include them.

SonarQube.Client/Api/V6_60/GetProjectBranchesRequest.cs
```diff
@@ -0,0 +1,58 @@
...
+        private class ServerProjectBranch
+        {
+            [JsonProperty("name")]
+            public string Name { get; set; }
+
+            [JsonProperty("isMain")]
+            public bool IsMain { get; set; }
+
+            [JsonProperty("analysisDate")]
+            public DateTimeOffset AnalysisDate { get; set; }
```
rita-g-sonarsource, line 55: misc and minor: maybe LastAnalysisDate  ?

SonarQube.Client.Tests/SonarQubeService_GetProjectBranchesAsync.cs
rita-g-sonarsource: The `type` looks worrisome, what other values can they return? and should we be filtering based on this?

rita-g-sonarsource (approved):

SonarQube.Client.Tests/SonarQubeService_GetProjectBranchesAsync.cs
duncanp-sonar: No idea, although looking at real data from SonarCloud and SonarQube, it may be the case that they aren't returning the same values.
However, for our current purposes we're only interested in whether the branch is the main branch or not. I'm making the (possibly unwarranted) assumption that `isMain` will only be true for one branch.

PR #92, opened 2021-01-29
Added secondary location hash updater

duncanp-sonar:
Relates to #1762

### DIFF
--- a/SonarQube.Client.Tests/Helpers/SecondaryIssueHashUpdaterTests.cs
+++ b/SonarQube.Client.Tests/Helpers/SecondaryIssueHashUpdaterTests.cs
@@ -1,0 +1,161 @@
+﻿/*
+ * SonarQube Client
+ * Copyright (C) 2016-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarQube.Client.Models;
+
+namespace SonarQube.Client.Helpers.Tests
+{
+    [TestClass]
+    public class SecondaryIssueHashUpdaterTests
+    {
+        private const string PrimaryIssueHash = "the primary issue hash should not be affected";
+        private readonly IChecksumCalculator checksumCalculator = new ChecksumCalculator();
+
+        [TestMethod]
+        public async Task Populate_NoIssues_NoOp()
+        {
+            var serviceMock = new Mock<ISonarQubeService>();
+
+            var testSubject = new SecondaryIssueHashUpdater();
+            await testSubject.UpdateHashesAsync(Array.Empty<SonarQubeIssue>(), serviceMock.Object, CancellationToken.None);
+
+            serviceMock.Invocations.Count.Should().Be(0);
+        }
+
+        [TestMethod]
+        public async Task Populate_NoSecondaryLocations_NoOp()
+        {
+            var serviceMock = new Mock<ISonarQubeService>();
+            var issues = new[]
+            {
+                CreateIssue("project1:key1", AddFlow()),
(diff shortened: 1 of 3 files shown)


SonarQube.Client/Models/SonarQubeIssue.cs
```diff
@@ -93,6 +93,12 @@ public IssueLocation(string filePath, string moduleKey, IssueTextRange textRange
         public string ModuleKey { get; }
         public IssueTextRange TextRange { get; }
         public string Message { get; }
+
+        /// <summary>
+        /// Note: currently the hash for secondary locations is calculated
+        /// post-construction, so we need to be able to update this property
+        /// </summary>
+        public string Hash { get; internal set; }
```
duncanp-sonar, line 101: As discussed previously, we could either update the location hash in-place, or build a completely new Issue-Flow-Location structure with the new data.
Updating in place is much simpler so I went with that.

SonarQube.Client/Helpers/SecondaryIssueHashUpdater.cs
```diff
@@ -0,0 +1,109 @@
...
+{
+    /// <summary>
+    /// Sets the hashes for any secondary locations in the supplied list of issues
+    /// </summary>
+    /// <remarks>
+    /// Currently secondary location hashes are not stored server-side, so we have to
+    /// calculate them ourselves. This means fetching the source code for each file
+    /// so we can get the line text and calculate the hash.
+    /// </remarks>
+    internal class SecondaryIssueHashUpdater
```
duncanp-sonar, line 38: I was originally going to split the functionality across multiple classes e.g. one which build the component->source map, one which used the map to set the hashes etc, but in the end I didn't bother as it just seemed to be adding unnecessary complexity for the amount of code involved.

SonarQube.Client/Helpers/SecondaryIssueHashUpdater.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        private static readonly IChecksumCalculator checksumCalculator = new ChecksumCalculator();
+        private Dictionary<string, string> moduleKeyToSourceMap;
+
+        internal async Task UpdateHashesAsync(IEnumerable<SonarQubeIssue> issues,
+            ISonarQubeService sonarQubeService,
+            CancellationToken cancellationToken)
+        {
+            var uniqueKeys = GetUniqueSecondaryLocationKeys(issues);
```
rita-g-sonarsource, line 47: Minor optimization: we can call `GetSecondaryLocations` only once and pass `IEnumerable<IssueLocation>` to GetUniqueSecondaryLocationKeys.

SonarQube.Client/Helpers/SecondaryIssueHashUpdater.cs
```diff
@@ -0,0 +1,107 @@
...
+                location.Hash = checksumCalculator.Calculate(firstLineOfIssue);
+            }
+        }
+
+        private string GetLineText(string moduleKey, int oneBasedLineNumber)
+        {
+            Debug.Assert(moduleKeyToSourceMap.ContainsKey(moduleKey), "Unexpected module key requested");
+
+            var source = moduleKeyToSourceMap[moduleKey];
+            var lines = source.Split('\n');
```
rita-g-sonarsource, line 97: Minor optimization: most secondary locations will be in the same file, in which case we would split the same content again and again. We can change `moduleKeyToSourceMap` to Dictionary<string, string[]> instead.

SonarQube.Client.Tests/Helpers/SecondaryIssueHashUpdaterTests.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            var testSubject = new SecondaryIssueHashUpdater();
+
+            // Act
+            await testSubject.UpdateHashesAsync(issues, serviceMock.Object, CancellationToken.None);
+
+            issues[0].Hash.Should().Be(PrimaryIssueHash);
+            issues[1].Hash.Should().Be(PrimaryIssueHash);
+
+            issues[0].Flows[0].Locations[0].Hash.Should().Be(checksumCalculator.Calculate(line1Contents));
```
rita-g-sonarsource, line 139: It's already an interface, why not mock it? its logic is irrelevant to this class, it shouldn't be a real dependency.
It can be easily replaced with a setup method:
```
SetupHash(checksumCalculator, line1Contents, "expected hash1");
SetupHash(checksumCalculator, line2Contents, "expected hash2");
...
issues[0].Flows[0].Locations[0].Hash.Should().Be("expected hash1");
issues[0].Flows[0].Locations[1].Hash.Should().Be("expected hash2");

private void SetupHash(Mock<IChecksumCalculator> checksumCalculator, string originalContent, string hashedContent) {...}
```

SonarQube.Client.Tests/Helpers/SecondaryIssueHashUpdaterTests.cs
```diff
@@ -0,0 +1,161 @@
...
+            serviceMock.Invocations.Count.Should().Be(0);
+        }
+
+        [TestMethod]
+        public async Task Populate_NoSecondaryLocations_NoOp()
+        {
+            var serviceMock = new Mock<ISonarQubeService>();
+            var issues = new[]
+            {
+                CreateIssue("project1:key1", AddFlow()),
```
rita-g-sonarsource, line 55: I don't think we can have a flow with no locations in reality.
Perhaps one of the issues can be with empty flows, to simulate the real conditions.

rita-g-sonarsource (approved):
Looks ok, just some comments

PR #95, opened 2021-01-29
Modified taint API to populate secondary location issues

### LINKED ISSUE #89: Modify GetIssues to optionally return line hashes for secondary locations

duncanp-sonar:
Fixes #89
Related to https://github.com/SonarSource/sonarlint-visualstudio/issues/1762

### DIFF
--- a/SonarQube.Client/SonarQubeService.cs
+++ b/SonarQube.Client/SonarQubeService.cs
@@ -41,6 +41,7 @@ namespace SonarQube.Client
         private readonly string userAgent;
         private readonly ILogger logger;
         private readonly IRequestFactorySelector requestFactorySelector;
+        private readonly ISecondaryIssueHashUpdater secondaryIssueHashUpdater;
 
         private HttpClient httpClient;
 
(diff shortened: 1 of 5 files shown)


SonarQube.Client/SonarQubeService.cs
```diff
@@ -323,13 +326,18 @@ public async Task<SonarQubeQualityProfile> GetQualityProfileAsync(string project
...
-            await InvokeRequestAsync<IGetTaintVulnerabilitiesRequest, SonarQubeIssue[]>(
+        public async Task<IList<SonarQubeIssue>> GetTaintVulnerabilitiesAsync(string projectKey, CancellationToken token)
+        {
+            var issues = await InvokeRequestAsync<IGetTaintVulnerabilitiesRequest, SonarQubeIssue[]>(
                 request =>
                 {
                     request.ProjectKey = projectKey;
                 }, token);
 
+            await secondaryIssueHashUpdater.UpdateHashesAsync(issues, this, token);
```
duncanp-sonar, line 337: This line is the only "real" change in the product code. Everything else is test-related plumbing.

rita-g-sonarsource (approved):

PR #101, opened 2021-05-07
Set IsConnected to false if ConnectAsync fails

duncanp-sonar:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/2406

### DIFF
--- a/SonarQube.Client.Tests/SonarQubeService_Lifecycle.cs
+++ b/SonarQube.Client.Tests/SonarQubeService_Lifecycle.cs
@@ -51,6 +51,49 @@ namespace SonarQube.Client.Tests
         }
 
         [TestMethod]
+        public async Task Connect_To_SonarQube_Invalid_Credentials()
+        {
+            // The earliest version that supports authentication
+            SetupRequest("api/server/version", "3.3.0.0");
+            SetupRequest("api/authentication/validate", "{ \"valid\": false }");
+
+            service.IsConnected.Should().BeFalse();
+            service.ServerInfo.Should().BeNull();
+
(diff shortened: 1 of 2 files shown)


SonarQube.Client.Tests/SonarQubeService_Lifecycle.cs
```diff
@@ -50,6 +50,49 @@ public async Task Connect_To_SonarQube_Valid_Credentials()
             service.ServerInfo.Version.Should().Be(new Version("3.3.0.0"));
         }
 
+        [TestMethod]
+        public async Task Connect_To_SonarQube_Invalid_Credentials()
```
duncanp-sonar, line 54: Added missing test case for invalid credentials.

SonarQube.Client/SonarQubeService.cs
```diff
@@ -136,28 +136,38 @@ public async Task ConnectAsync(ConnectionInformation connection, CancellationTok
...
-            var credentialResponse = await InvokeRequestAsync<IValidateCredentialsRequest, bool>(token);
-            if (!credentialResponse)
+                logger.Debug($"Credentials accepted.");
+            }
+            catch
             {
                 IsConnected = false;
                 ServerInfo = null;
-                throw new InvalidOperationException("Invalid credentials");
+                throw;
```
duncanp-sonar, line 169: I don't want to change the existing behaviour of allowing the exceptions to escape since the callers might be relying on it.

All we can do is make sure the "is connected" state is set correctly.

rita-sonarsource (approved):

PR #105, opened 2021-11-15
Fix SC dashboard URL

rita-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/2753

### DIFF
--- a/SonarQube.Client/SonarQubeService.cs
+++ b/SonarQube.Client/SonarQubeService.cs
@@ -234,9 +234,12 @@ namespace SonarQube.Client
         {
             EnsureIsConnected();
 
-            const string ProjectDashboardRelativeUrl = "dashboard/index/{0}";
+            const string SonarQube_ProjectDashboardRelativeUrl = "dashboard/index/{0}";
+            const string SonarCloud_ProjectDashboardRelativeUrl = "project/overview?id={0}";
 
-            return new Uri(httpClient.BaseAddress, string.Format(ProjectDashboardRelativeUrl, projectKey));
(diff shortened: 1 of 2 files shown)


SonarQube.Client/SonarQubeService.cs
```diff
@@ -234,9 +234,12 @@ public Uri GetProjectDashboardUrl(string projectKey)
         {
             EnsureIsConnected();
 
-            const string ProjectDashboardRelativeUrl = "dashboard/index/{0}";
+            const string SonarQube_ProjectDashboardRelativeUrl = "dashboard/index/{0}";
```
rita-sonarsource, line 237: This is what we do for hotspots as well.

I originally created a request for this (see https://github.com/SonarSource/sonarqube-webclient-dotnet/commit/d2a8df353c0e3433091b21f4b922c57042d6874f) but it was messy integrating it into SLVS -- this code is used by the WPF commands and using Task.FromResult didn't feel right when looking at all these changes combined.

rita-sonarsource: FYI, SQ's URL was also changed, to `dashboard?id={0}` but there is a redirect so the old one still works. I left it as is for backwards compatibility with older SQ versions.

duncanp-sonar (approved):

SonarQube.Client/SonarQubeService.cs
duncanp-sonar: That seems like a more sensible approach...

PR #116, opened 2022-06-21
Implement Equals on ServerExclusions

rita-gorokhod:
Prep for https://github.com/SonarSource/sonarlint-visualstudio/issues/3083

### DIFF
--- a/SonarQube.Client/Models/ServerExclusions.cs
+++ b/SonarQube.Client/Models/ServerExclusions.cs
@@ -65,8 +65,15 @@ namespace SonarQube.Client.Models
 
         public bool Equals(ServerExclusions other)
         {
-            if (ReferenceEquals(null, other)) return false;
-            if (ReferenceEquals(this, other)) return true;
+            if (other == null)
+            {
+                return false;
+            }
+
+            if (ReferenceEquals(this, other))
+            {
+                return true;
+            }
 
             return ToString().Equals(other.ToString());
         }


SonarQube.Client/Models/ServerExclusions.cs
```diff
@@ -57,5 +57,23 @@ public override string ToString()
...
+        {
+            if (ReferenceEquals(null, other)) return false;
+            if (ReferenceEquals(this, other)) return true;
+
+            return ToString().Equals(other.ToString());
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
```
duncanp-sonar, line 76: This is doing nothing in a confusing way.
We should provide a correct implementation or nothing (and we should provide a correct implementation rather than nothing) e.g. just return `this.ToString().GetHashCode()`;

duncanp-sonar: Sorry, missed that this was still a draft.

rita-sonarsource: Yep I'm adding it now 👍

SonarQube.Client/Models/ServerExclusions.cs
```diff
@@ -57,5 +57,23 @@ public override string ToString()
                    "\n    Exclusions: " + string.Join(",", Exclusions) +
                    "\n    Global Exclusions: " + string.Join(",", GlobalExclusions);
         }
+
+        public override bool Equals(object obj)
+        {
+            return Equals((ServerExclusions)obj);
```
duncanp-sonar, line 63: This needs be a soft cast, otherwise this will throw if the object isn't a `ServerExclusions` instance.

rita-sonarsource: Just failed on this... Fixed now, my bad.

SonarQube.Client/Models/ServerExclusions.cs
```diff
@@ -57,5 +57,23 @@ public override string ToString()
...
+        }
+
+        public bool Equals(ServerExclusions other)
+        {
+            if (ReferenceEquals(null, other)) return false;
```
duncanp-sonar, line 68: `if (other == null)` would be easier to follow.

SonarQube.Client/Models/ServerExclusions.cs
```diff
@@ -57,5 +57,23 @@ public override string ToString()
...
+            if (ReferenceEquals(this, other)) return true;
```
duncanp-sonar, line 69: I think we always use curly-brackets, even for single-line statements as simple as this one.

duncanp-sonar (commented):
Couple of minor comments, otherwise looks good.

ugras-ergun-sonarsource (approved):

PR #125, opened 2022-07-13
Merge branch-exclusions

rita-sonarsource:


duncanp-sonar:
@rita-sonarsource I was worried about this earlier when I was merging the mend branch and updating the submodule. However, I thought that I had managed to successfully include the exclusions commit i.e. it looks like they are already in `master`: https://github.com/SonarSource/sonarqube-webclient-dotnet/commits/master

image

Or did I still manage to mess it up somehow?

PR #126, opened 2022-07-14
Add js/ts plugin information

rita-sonarsource:


SonarQube.Client/Models/SonarQubeLanguage.cs
```diff
@@ -26,6 +26,8 @@ public class SonarQubeLanguage
         public static readonly SonarQubeLanguage VbNet = new SonarQubeLanguage("vbnet", "VB.NET", "SonarVB");
         public static readonly SonarQubeLanguage Cpp = new SonarQubeLanguage("cpp", "C++", "SonarCFamily");
         public static readonly SonarQubeLanguage C = new SonarQubeLanguage("c", "C", "SonarCFamily");
+        public static readonly SonarQubeLanguage Js = new SonarQubeLanguage("js", "JavaScript", "javascript");
```
rita-sonarsource, line 29: This is used in SLVS language class https://github.com/SonarSource/sonarlint-visualstudio/blob/master/src/Core/Language.cs#L43

I don't think it should've been here in this module. I can remove the static declarations and leave just the type here, moving the statics to SLVS. WDYT?

PR #129, opened 2022-08-03
Mark service as Connected only after full initialization

rita-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/696

### DIFF
--- a/SonarQube.Client/SonarQubeService.cs
+++ b/SonarQube.Client/SonarQubeService.cs
@@ -54,7 +54,7 @@ namespace SonarQube.Client
             return await Task.FromResult<bool>(hasOrganisations);
         }
 
-        public bool IsConnected { get; private set; }
+        public bool IsConnected => ServerInfo != null;
 
         public ServerInfo ServerInfo { get; private set; }
 
@@ -103,12 +103,18 @@ namespace SonarQube.Client
         /// <param name="configure">Action that configures a type instance that implements TRequest.</param>
         /// <param name="token">Cancellation token.</param>
         /// <returns>Returns the result of the request invocation.</returns>
-        protected virtual async Task<TResponse> InvokeRequestAsync<TRequest, TResponse>(Action<TRequest> configure,
+        protected virtual Task<TResponse> InvokeRequestAsync<TRequest, TResponse>(Action<TRequest> configure,
             CancellationToken token)
             where TRequest : IRequest<TResponse>
         {
             EnsureIsConnected();
 
+            return InvokeUnauthenticatedRequestAsync<TRequest, TResponse>(configure, token);
+        }
+
+        private async Task<TResponse> InvokeUnauthenticatedRequestAsync<TRequest, TResponse>(Action<TRequest> configure, CancellationToken token)
+            where TRequest : IRequest<TResponse>
+        {
             var request = requestFactory.Create<TRequest>(ServerInfo);
             configure(request);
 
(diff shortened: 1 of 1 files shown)


SonarQube.Client/SonarQubeService.cs
```diff
@@ -138,33 +144,29 @@ public async Task ConnectAsync(ConnectionInformation connection, CancellationTok
...
+                var versionResponse = await InvokeUnauthenticatedRequestAsync<IGetVersionRequest, string>(request => { }, token);
+                var serverInfo = new ServerInfo(Version.Parse(versionResponse), connection.IsSonarCloud ? ServerType.SonarCloud : ServerType.SonarQube);
 
-                logger.Info($"Connected to {serverTypeDescription} '{ServerInfo.Version}'.");
+                logger.Info($"Connected to {serverTypeDescription} '{serverInfo.Version}'.");
 
                 logger.Debug($"Validating the credentials...");
 
-                var credentialResponse = await InvokeRequestAsync<IValidateCredentialsRequest, bool>(token);
+                var credentialResponse = await InvokeUnauthenticatedRequestAsync<IValidateCredentialsRequest, bool>(request => { }, token);
```
rita-sonarsource, line 158: I feel that `InvokeUnauthenticatedRequestAsync` is not the best name for it.
It's not unauthenticated, it's more "InvokeWithoutWaitingForConnection".

SonarQube.Client/SonarQubeService.cs
```diff
@@ -54,7 +54,7 @@ public async Task<bool> HasOrganizations(CancellationToken token)
             return await Task.FromResult<bool>(hasOrganisations);
         }
 
-        public bool IsConnected { get; private set; }
+        public bool IsConnected => ServerInfo != null;
```
rita-sonarsource, line 57: There was no need to have an additional boolean, ServerInfo serves the same purpose.

SonarQube.Client/SonarQubeService.cs
duncanp-sonar: `InvokeUncheckedRequestAsync`? And maybe renaming the other method `InvokeCheckedRequestAsync`.

Or some variation thereof: `InvokeRequestWithCheckAsync`.

`Checked` still isn't great, but it's slightly better than `Unauthenticated`.

SonarQube.Client/SonarQubeService.cs
```diff
@@ -103,12 +103,18 @@ public SonarQubeService(HttpMessageHandler messageHandler, string userAgent, ILo
...
+        protected virtual Task<TResponse> InvokeRequestAsync<TRequest, TResponse>(Action<TRequest> configure,
             CancellationToken token)
             where TRequest : IRequest<TResponse>
         {
             EnsureIsConnected();
 
+            return InvokeUnauthenticatedRequestAsync<TRequest, TResponse>(configure, token);
+        }
+
+        private async Task<TResponse> InvokeUnauthenticatedRequestAsync<TRequest, TResponse>(Action<TRequest> configure, CancellationToken token)
```
duncanp-sonar, line 115: Deserves a comment e.g.

> Executes the call without checking whether the connection to the server has been established. This should only normally be used directly while connecting.
Other uses should call `InvokeCheckedRequestAsync`.

SonarQube.Client/SonarQubeService.cs
```diff
@@ -103,12 +103,18 @@ public SonarQubeService(HttpMessageHandler messageHandler, string userAgent, ILo
         /// <param name="configure">Action that configures a type instance that implements TRequest.</param>
         /// <param name="token">Cancellation token.</param>
         /// <returns>Returns the result of the request invocation.</returns>
-        protected virtual async Task<TResponse> InvokeRequestAsync<TRequest, TResponse>(Action<TRequest> configure,
+        protected virtual Task<TResponse> InvokeRequestAsync<TRequest, TResponse>(Action<TRequest> configure,
```
duncanp-sonar, line 106: The `protected virtual` was added so the `MEFSonarQubeService` could override make sure all the web calls happened on the background thread. The override was here, because this was the low-level method.

It isn't anymore, so this method can go back to being `private`, and the new method can be `protected virtual`.

duncanp-sonar (approved):

SonarQube.Client/SonarQubeService.cs
```diff
@@ -103,12 +103,22 @@ public SonarQubeService(HttpMessageHandler messageHandler, string userAgent, ILo
...
             EnsureIsConnected();
 
+            return InvokeUncheckedRequestAsync<TRequest, TResponse>(configure, token);
+        }
+
+        /// <summary>
+        /// Executes the call without checking whether the connection to the server has been established. This should only normally be used directly while connecting.
+        /// Other uses should call <see cref="InvokeCheckedRequestAsync{TRequest,TResponse}(System.Threading.CancellationToken)"/>.
+        /// </summary>
+        protected virtual async Task<TResponse> InvokeUncheckedRequestAsync<TRequest, TResponse>(Action<TRequest> configure, CancellationToken token)
```
rita-sonarsource, line 119: @duncanp-sonar you meant changing this method to virtual, right?

duncanp-sonar: Yep; it's now the inner-most call that actually does the work.

PR #132, opened 2022-12-05
Add branch parameter to GetIssuesAsync

duncanp-sonar:
Fixes SLVS#58

### DIFF
--- a/SonarQube.Client.Tests/Infra/MocksHelper.cs
+++ b/SonarQube.Client.Tests/Infra/MocksHelper.cs
@@ -32,6 +32,26 @@ namespace SonarQube.Client.Tests.Infra
     {
         public const string ValidBaseAddress = "http://localhost";
 
+        /// <summary>
+        /// Sets up the HTTP message handler mock to respond to any request string 
+        /// </summary>
+        public static void SetupHttpRequest(Mock<HttpMessageHandler> messageHandlerMock, string response,
+            HttpStatusCode statusCode = HttpStatusCode.OK)
+        {
+            messageHandlerMock.Protected()
+                .Setup<Task<HttpResponseMessage>>("SendAsync",
+                    ItExpr.IsAny<HttpRequestMessage>(),
+                    ItExpr.IsAny<CancellationToken>())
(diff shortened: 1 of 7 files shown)


SonarQube.Client.Tests/Infra/MocksHelper.cs
```diff
@@ -32,6 +32,26 @@ internal static class MocksHelper
     {
         public const string ValidBaseAddress = "http://localhost";
 
+        /// <summary>
+        /// Sets up the HTTP message handler mock to respond to any request string 
+        /// </summary>
+        public static void SetupHttpRequest(Mock<HttpMessageHandler> messageHandlerMock, string response,
```
duncanp-sonar, line 38: New helper method that doesn't specify the expected "request" string.

If we just want to write a method that checks the expected request string has been passed then this new method is more useful than the existing one in this class.

SonarQube.Client.Tests/Requests/Api/V7_20/GetIssuesRequestWrapperTests.cs
```diff
@@ -0,0 +1,91 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using SonarQube.Client.Api.V7_20;
+using SonarQube.Client.Tests.Infra;
+using static SonarQube.Client.Tests.Infra.MocksHelper;
+
+namespace SonarQube.Client.Tests.Requests.Api.V7_20
+{
+    [TestClass]
+    public class GetIssuesRequestWrapperTests
```
duncanp-sonar, line 36: `GetIssuesRequestWrapper` is a helper class that internally makes three requests to `GetIssues`, first for code smells, then for bugs and finally for vulnerabilities.
This was a hacky way to partially work round a limitation in the SQ/SC API that meant a request could only return a maximum of 10000 results.

Previously, there weren't any direct tests for this class, although it was tested indirectly by the "GetSuppressedIssuesAsync" tests.

I've added one simple test to make sure that three requests are made and the expected query parameters are set.

ugras-ergun-sonarsource (approved):

PR #142, opened 2023-01-12
Add IGetSonarLintEventStream implementation

rita-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3528

### DIFF
--- a/SonarQube.Client.Tests/Infra/MocksHelper.cs
+++ b/SonarQube.Client.Tests/Infra/MocksHelper.cs
@@ -19,10 +19,10 @@
  */
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Threading;
 using System.Threading.Tasks;
 using Moq;
@@ -59,17 +59,30 @@ namespace SonarQube.Client.Tests.Infra
         public static void SetupHttpRequest(Mock<HttpMessageHandler> messageHandlerMock, string requestRelativePath, string response,
             HttpStatusCode statusCode = HttpStatusCode.OK, string basePath = ValidBaseAddress)
         {
+            var responseMessage = new HttpResponseMessage
+            {
+                StatusCode = statusCode,
+                Content = new StringContent(response)
+            };
+
+            SetupHttpRequest(messageHandlerMock, requestRelativePath, responseMessage, basePath);
+        }
+
+        public static void SetupHttpRequest(Mock<HttpMessageHandler> messageHandlerMock,
+            string requestRelativePath, 
+            HttpResponseMessage responseMessage, 
+            string basePath = ValidBaseAddress,
+            params MediaTypeHeaderValue[] headers)
+        {
             var baseUri = new Uri(basePath);
+
             messageHandlerMock.Protected()
                 .Setup<Task<HttpResponseMessage>>("SendAsync",
                     ItExpr.Is<HttpRequestMessage>(m =>
-                        m.RequestUri == new Uri(baseUri, requestRelativePath)),
+                        m.RequestUri == new Uri(baseUri, requestRelativePath) &&
+                        headers.All(header => m.Headers.Accept.Contains(header))),
                     ItExpr.IsAny<CancellationToken>())
-                .Returns(Task.FromResult(new HttpResponseMessage
-                {
-                    StatusCode = statusCode,
-                    Content = new StringContent(response)
-                }));
+                .Returns(Task.FromResult(responseMessage));
         }
 
         /// <summary>
--- a/SonarQube.Client.Tests/Requests/Api/V9_4/PushRequestTests.cs
+++ b/SonarQube.Client.Tests/Requests/Api/V9_4/PushRequestTests.cs
@@ -28,7 +28,6 @@ using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
-using Org.BouncyCastle.Crypto.IO;
 using SonarQube.Client.Api.V9_4;
 using SonarQube.Client.Logging;
 using SonarQube.Client.Tests.Infra;
@@ -42,8 +41,8 @@ namespace SonarQube.Client.Tests.Requests.Api.V9_4
         public async Task InvokeAsync_ReturnsCorrectStream()
         {
             using var testedStream = new MemoryStream(Encoding.UTF8.GetBytes("hello this is a test"));
-
             var messageHandler = new Mock<HttpMessageHandler>();
+            using var httpClient = new HttpClient(messageHandler.Object) { BaseAddress = new Uri("http://localhost") };
 
             MocksHelper.SetupHttpRequest(
                 messageHandler,
@@ -51,19 +50,15 @@ namespace SonarQube.Client.Tests.Requests.Api.V9_4
                 responseMessage: new HttpResponseMessage {Content = new StreamContent(testedStream) },
                 headers: MediaTypeHeaderValue.Parse("text/event-stream"));
 
-            using var httpClient = new HttpClient(messageHandler.Object) {BaseAddress = new Uri("http://localhost")};
-
             var testSubject = new PushRequest {Languages = "somelang", ProjectKey = "someproj", Logger = Mock.Of<ILogger>()};
 
             using var response = await testSubject.InvokeAsync(httpClient, CancellationToken.None);
             response.Should().NotBeNull();
+            messageHandler.VerifyAll();
 
             using var reader = new StreamReader(response, Encoding.UTF8);
             var responseString = await reader.ReadToEndAsync();
-
             responseString.Should().Be("hello this is a test");
-
-            messageHandler.VerifyAll();
         }
     }
 }
--- a/SonarQube.Client.Tests/Requests/Api/V9_4/PushRequestTests.cs
+++ b/SonarQube.Client.Tests/Requests/Api/V9_4/PushRequestTests.cs
@@ -46,11 +46,11 @@ namespace SonarQube.Client.Tests.Requests.Api.V9_4
 
             MocksHelper.SetupHttpRequest(
                 messageHandler,
-                requestRelativePath: "api/push/sonarlint_events?languages=somelang&projectKeys=someproj",
+                requestRelativePath: "api/push/sonarlint_events?languages=cs%2Cvbnet%2Ccpp%2Cc%2Cjs%2Cts&projectKeys=someproj",
                 responseMessage: new HttpResponseMessage {Content = new StreamContent(testedStream) },
                 headers: MediaTypeHeaderValue.Parse("text/event-stream"));
 
-            var testSubject = new PushRequest {Languages = "somelang", ProjectKey = "someproj", Logger = Mock.Of<ILogger>()};
+            var testSubject = new PushRequest {ProjectKey = "someproj", Logger = Mock.Of<ILogger>()};
 
             using var response = await testSubject.InvokeAsync(httpClient, CancellationToken.None);
             response.Should().NotBeNull();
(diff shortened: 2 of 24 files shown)


SonarQube.Client.Tests/Infra/MocksHelper.cs
```diff
@@ -58,18 +58,31 @@ internal static class MocksHelper
...
+            var responseMessage = new HttpResponseMessage
+            {
+                StatusCode = statusCode,
+                Content = new StringContent(response)
+            };
+
+            SetupHttpRequest(messageHandlerMock, requestRelativePath, responseMessage, basePath);
+        }
+
+        public static void SetupHttpRequest(Mock<HttpMessageHandler> messageHandlerMock,
```
rita-sonarsource, line 71: New overload that accepts HttpResponseMessage instead of a string, and the optional additional headers.
The existing overload calls this one.

SonarQube.Client.Tests/SonarQube.Client.Tests.csproj
```diff
@@ -9,6 +9,8 @@
     <RootNamespace>SonarQube.Client.Tests</RootNamespace>
 
     <RequiresNewtonsoft>true</RequiresNewtonsoft>
+
```
rita-sonarsource, line 12: So I could use in-line "using"

SonarQube.Client.Tests/SonarQubeService_CreateServerSentEventsSession.cs
```diff
@@ -0,0 +1,83 @@
...
+        public async Task CreateServerSentEventsSession_UnsupportedServerVersion_InvalidOperationException()
+        {
+            await ConnectToSonarQube("3.3.0.0");
+
+            Func<Task<IServerSentEventsSession>> func = async () => await service.CreateServerSentEventsSession("myProject", CancellationToken.None);
+
+            const string expectedErrorMessage =
+                "Could not find compatible implementation of 'IPushRequest' for SonarQube 3.3.0.0.";
+
+            func.Should().ThrowExactly<InvalidOperationException>().WithMessage(expectedErrorMessage);
```
rita-sonarsource, line 64: This is a bit ugly, the client will have to handle unsupported SQ versions and we return "InvalidOperationException". NotImplemented or UnsupportedMethod would've been more suitable IMO, but that's InvalidOperationException is what the infra currently returns.

SonarQube.Client/Api/V9_4/PushRequest.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        protected override string Path => "api/push/sonarlint_events";
+
+        protected override MediaTypeWithQualityHeaderValue[] Headers =>
+            new[]
+            {
+                MediaTypeWithQualityHeaderValue.Parse("text/event-stream")
+            };
+
+        protected override async Task<Result<Stream>> ReadResponseAsync(HttpResponseMessage httpResponse)
```
rita-sonarsource, line 43: The base implementation also calls the abstract ParseResponse, which we no longer need. I kept the changes to a minimum.

SonarQube.Client/Models/ServerSentEvents/IServerSentEventsSession.cs
```diff
@@ -18,13 +18,30 @@
...
 {
+    /// <summary>
+    /// Wraps the stream response from the server and reads from it
+    /// </summary>
     public interface IServerSentEventsSession
     {
         Task<IServerEvent> ReadAsync();
     }
+
+    internal class ServerSentEventsSession : IServerSentEventsSession
```
rita-sonarsource, line 36: Will be implemented in future PRs

SonarQube.Client/Requests/RequestBase.cs
```diff
@@ -82,12 +86,17 @@ protected async Task<Result<TResponse>> InvokeUncheckedAsync(HttpClient httpClie
...
             Logger.Debug("Sending Http request:");
 
-            string query = QueryStringSerializer.ToQueryString(this);
+            var query = QueryStringSerializer.ToQueryString(this);
 
             var pathAndQuery = string.IsNullOrEmpty(query) ? Path : $"{Path}?{query}";
 
             var httpRequest = new HttpRequestMessage(HttpMethod, new Uri(pathAndQuery, UriKind.Relative));
 
+            foreach (var header in Headers)
```
rita-sonarsource, line 95: Needed to support `Accept: text/event-stream` header

SonarQube.Client/Api/V9_4/PushRequest.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        protected override Stream ParseResponse(string response)
+        {
+            // should not be called
+            throw new InvalidOperationException();
+        }
+
+        [JsonProperty("languages")]
+        public string Languages { get; set; } = string.Join(",", SonarQubeLanguage.AllLanguages.Select(x=> x.Key));
```
georgii-borovinskikh-sonarsource, line 57: Cache this?

SonarQube.Client/Api/V9_4/PushRequest.cs
```diff
@@ -0,0 +1,62 @@
...
+            };
+
+        protected override async Task<Result<Stream>> ReadResponseAsync(HttpResponseMessage httpResponse)
+        {
+            var stream = await httpResponse.Content.ReadAsStreamAsync();
+
+            return new Result<Stream>(httpResponse, stream);
+        }
+
+        protected override Stream ParseResponse(string response)
```
georgii-borovinskikh-sonarsource, line 50: why does this accept a string but them return a stream?

SonarQube.Client.Tests/SonarQube.Client.Tests.csproj
```diff
@@ -9,6 +9,8 @@
...
+    <LangVersion>latest</LangVersion>
```
georgii-borovinskikh-sonarsource, line 13: use a fixed version instead?

SonarQube.Client.Tests/Requests/Api/V9_4/PushRequestTests.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace SonarQube.Client.Tests.Requests.Api.V9_4
+{
+    [TestClass]
+    public class PushRequestTests
+    {
+        [TestMethod]
+        public async Task InvokeAsync_ReturnsCorrectStream()
+        {
+            using var testedStream = new MemoryStream(Encoding.UTF8.GetBytes("hello this is a test"));
+
```
georgii-borovinskikh-sonarsource, line 45: Empty lines, discussed in person

SonarQube.Client.Tests/SonarQubeService_CreateServerSentEventsSession.cs
georgii-borovinskikh-sonarsource: Maybe use Result<IServerSentEventsSession> with enum flags for error types instead of Exceptions?

georgii-borovinskikh-sonarsource (changes requested):

SonarQube.Client.Tests/SonarQubeService_CreateServerSentEventsSession.cs
rita-sonarsource: I can't change it right now as it would break everything, but in the future we should probably handle this

SonarQube.Client/Api/V9_4/PushRequest.cs
rita-sonarsource: Part of the abstract base method, it will not be called. The infrastructure classes assume that you will always get a json and always deserialize it to something.

SonarQube.Client/Api/IPushRequest.cs
```diff
@@ -0,0 +1,31 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System.IO;
+using SonarQube.Client.Requests;
+
+namespace SonarQube.Client.Api
+{
+    internal interface IPushRequest : IRequest<Stream>
```
duncanp-sonar, line 26: Missing doc comment.

The name isn't very descriptive. Something like `IPushSonarLintEvents`, or `IGetSonarLintEventStream` would be clearer.

SonarQube.Client/Api/V9_4/PushRequest.cs
georgii-borovinskikh-sonarsource: Right, I missed this, my bad

SonarQube.Client/Requests/RequestBase.cs
```diff
@@ -47,6 +48,9 @@ public abstract class RequestBase<TResponse> : IRequest<TResponse>
         [JsonIgnore]
         protected virtual HttpMethod HttpMethod => HttpMethod.Get;
 
+        [JsonIgnore]
+        protected virtual MediaTypeWithQualityHeaderValue[] Headers => Array.Empty<MediaTypeWithQualityHeaderValue>();
```
duncanp-sonar, line 52: `Headers` is a bit general. `AllowedMediaTypeHeaders` might be better.

georgii-borovinskikh-sonarsource (approved):

SonarQube.Client/Api/DefaultConfiguration.cs
```diff
@@ -54,7 +54,8 @@ public static RequestFactory ConfigureSonarQube(RequestFactory requestFactory)
                 .RegisterRequest<IGetIssuesRequest, V7_20.GetIssuesRequestWrapper>("7.2")
                 .RegisterRequest<IGetHotspotRequest, V8_6.GetHotspotRequest>("8.6")
                 .RegisterRequest<IGetTaintVulnerabilitiesRequest, V8_6.GetTaintVulnerabilitiesRequest>("8.6")
-                .RegisterRequest<IGetExclusionsRequest, V7_20.GetExclusionsRequest>("7.2");
+                .RegisterRequest<IGetExclusionsRequest, V7_20.GetExclusionsRequest>("7.2")
+                .RegisterRequest<IPushRequest, V9_4.PushRequest>("9.4");
```
duncanp-sonar, line 58: How are we going to handle the new API method call being made against SonarCloud?
Currently, I think the SC request factory will throw.

Is it the responsibility of the calling client to not call the new method if they are connected to SonarCloud, or the responsibility of this webclient layer to handle it (e.g. by returning a no-op response of some kind e.g. an empty/null stream)?

I think it would be preferable if the callers didn't need to know about differences between the SonarQube and SonarCloud.

duncanp-sonar (commented):
A few questions/comments inline.

SonarQube.Client/Api/IPushRequest.cs
rita-sonarsource: Changed to IGetSonarLintEventStream

SonarQube.Client/Api/DefaultConfiguration.cs
rita-sonarsource: It would be the same as an unsupported old SQ version: the API will throw InvalidOperationException. 
It will then be up to the caller to know that we are not connected to a supported instance. This is what I meant in this comment https://github.com/SonarSource/sonarqube-webclient-dotnet/pull/142#discussion_r1067902116: 
`This is a bit ugly, the client will have to handle unsupported SQ versions and we return "InvalidOperationException". NotImplemented or UnsupportedMethod would've been more suitable IMO, but that's InvalidOperationException is what the infra currently returns.`

It would've been much better to return something else, i.e. NotSupportedException, but we can't change it without changing the infra of all the requests.

duncanp-sonar: I don't think the type of the exception that is thrown is necessarily the problem; we're still throwing something the client needs to handle.

I was suggesting that we could add a dummy request to the SonarCloud factory so the request doesn't fail, but instead returns a no-op e.g. a [null-object](https://en.wikipedia.org/wiki/Null_object_pattern). That way the caller wouldn't need to know about differences between SC and SQ.

It would mean the client would create instances of classes that are waiting for events that will never come, but the overhead of that is minimal.

rita-sonarsource: as we discussed offline, for now I'll leave it to SLVS

PR #143, opened 2023-01-12
Update .net framework to 472

rita-sonarsource:
Prep for using channels nuget package

### DIFF
--- a/SonarQube.Client.Tests/SonarQubeService_GetNotificationEventsAsync.cs
+++ b/SonarQube.Client.Tests/SonarQubeService_GetNotificationEventsAsync.cs
@@ -39,7 +39,7 @@ namespace SonarQube.Client.Tests
             var now = new DateTimeOffset(2017, 10, 19, 13, 0, 0, TimeSpan.FromHours(2));
 
             await ConnectToSonarQube("6.6.0.0");
-            SetupRequest("api/developers/search_events?projects=my_project&from=2017-10-19T13:00:00%2B0200",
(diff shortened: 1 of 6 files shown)


SonarQube.Client.Tests/SonarQubeService_GetNotificationEventsAsync.cs
```diff
@@ -39,7 +39,7 @@ public async Task GetNotifications_ExampleFromSonarQube()
             var now = new DateTimeOffset(2017, 10, 19, 13, 0, 0, TimeSpan.FromHours(2));
 
             await ConnectToSonarQube("6.6.0.0");
-            SetupRequest("api/developers/search_events?projects=my_project&from=2017-10-19T13:00:00%2B0200",
+            SetupRequest("api/developers/search_events?projects=my_project&from=2017-10-19T13%3A00%3A00%2B0200",
```
rita-sonarsource, line 42: `:` was replaced with `%3A` after upgrading the package 🤷‍♀️

georgii-borovinskikh-sonarsource: Revert? I don't think we needed to change it

georgii-borovinskikh-sonarsource (approved):

SonarQube.Client.Tests/SonarQubeService_GetNotificationEventsAsync.cs
rita-sonarsource: We do, it fails otherwise

duncanp-sonar (approved):

PR #145, opened 2023-01-26
remove server contract abstraction, add the minimal set of properties that are needed for filtering

georgii-borovinskikh-sonarsource:


SonarQube.Client/Models/ServerSentEvents/ClientContract/ITaintServerEvent.cs
```diff
@@ -18,23 +18,24 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-using SonarQube.Client.Models.ServerSentEvents.ServerContract;
-
 namespace SonarQube.Client.Models.ServerSentEvents.ClientContract
 {
     public interface ITaintServerEvent : IServerEvent
     {
+        string ProjectKey { get; }
```
rita-sonarsource, line 25: Do these fields exist on ITaintVulnerabilityClosedServerEvent ? 
I think it would be better to put these fields on ITaintVulnerabilityRaisedServerEvent for readability.

rita-sonarsource (approved):
LGTM

SonarQube.Client/Models/ServerSentEvents/ClientContract/ITaintServerEvent.cs
georgii-borovinskikh-sonarsource: Yes, they were. You can see it in the diff: SonarQube.Client/Models/ServerSentEvents/ServerContract/ITaintVulnerabilityClosedSqServerEvent.cs

PR #147, opened 2023-01-30
add implementations for basic client models

georgii-borovinskikh-sonarsource:


SonarQube.Client/Models/ServerSentEvents/ClientContract/IIssueChangedServerEvent.cs
```diff
@@ -33,6 +33,20 @@ public interface IIssueChangedServerEvent : IServerEvent
         // also has Severity and Type that we don't care about
     }
 
+    public class IssueChangedServerEvent : IIssueChangedServerEvent
+    {
+        public IssueChangedServerEvent(string projectKey, bool isResolved, IBranchAndIssueKey[] branchAndIssueKeys)
```
rita-sonarsource, line 38: I was hoping to deserialize the entire string into this, so perhaps instead of a ctor we need Setters

georgii-borovinskikh-sonarsource: WDYM? Can't you do it with the ctor as well? Newtonsoft is smart enough to support that

rita-sonarsource: hmm good question, if it works then great 👍 
I'd then add `?? throw new ArgumentNullException` + tests, this way we don't need to worry about invalid data.

PR #151, opened 2023-02-02
Aggregate event lines into ISqServerEvent

rita-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3606


SonarQube.Client/Models/ServerSentEvents/SSEStreamWriter.cs
```diff
@@ -35,39 +37,60 @@ public interface ISSEStreamWriter : IDisposable
...
+            var eventLines = new List<string>();
+
+            while (!serverStreamReader.EndOfStream && !cancellationToken.IsCancellationRequested)
+            {
+                var line = await serverStreamReader.ReadLineAsync();
+                var isEventEnd = line == string.Empty;
+
+                if (isEventEnd)
+                {
+                    var parsedEvent = sqServerSentEventParser.Parse(eventLines.ToList());
```
rita-sonarsource, line 80: I don't understand why, but if I don't pass `eventLines.ToList()`, sqServerSentEventParser's mock invocations fail that it expected the wrong thing.

SonarQube.Client/Models/ServerSentEvents/SSEStreamWriter.cs
```diff
@@ -35,39 +37,60 @@ public interface ISSEStreamWriter : IDisposable
...
+        public async Task BeginListening()
         {
-            // todo: next PR
-            throw new NotImplementedException();
+            var eventLines = new List<string>();
+
+            while (!serverStreamReader.EndOfStream && !cancellationToken.IsCancellationRequested)
+            {
+                var line = await serverStreamReader.ReadLineAsync();
+                var isEventEnd = line == string.Empty;
```
georgii-borovinskikh-sonarsource, line 76: use ReferenceEquals explicitly?

SonarQube.Client/Models/ServerSentEvents/SSEStreamWriter.cs
georgii-borovinskikh-sonarsource: Allocating an extra list is not good. Fixing or even removing the test is better

SonarQube.Client/Models/ServerSentEvents/SSEStreamWriter.cs
```diff
@@ -35,39 +37,60 @@ public interface ISSEStreamWriter : IDisposable
...
     }
 
+    /// <summary>
+    /// Aggregates stream lines into events.
+    /// Code on the java side: https://github.com/SonarSource/sonarlint-core/blob/171ca4d75c24033e115a81bd7481427cd1f39f4c/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/stream/EventBuffer.java
+    /// </summary>
     internal class SSEStreamWriter : ISSEStreamWriter
     {
-        private readonly Stream serverStream;
+        private readonly StreamReader serverStreamReader;
```
georgii-borovinskikh-sonarsource, line 46: Calling this networkStreamReader would be less confusing IMO

SonarQube.Client.Tests/Models/ServerSentEvents/SSEStreamWriterTests.cs
```diff
@@ -0,0 +1,146 @@
...
+            sqServerSentEventParser.VerifyAll();
+            sqServerSentEventParser.VerifyNoOtherCalls();
+        }
+
+        [TestMethod]
+        public void Dispose_ClosesStream()
+        {
+            var streamReader = CreateStreamReader(content: "content");
+
+            var testSubject = CreateTestSubject(streamReader, sqEventsChannel: CreateChannel());
```
georgii-borovinskikh-sonarsource, line 107: why pass optional here?

SonarQube.Client.Tests/Models/ServerSentEvents/SSEStreamWriterTests.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        private static Channel<ISqServerEvent> CreateChannel() => Channel.CreateUnbounded<ISqServerEvent>();
+
+        private static SSEStreamWriter CreateTestSubject(
+            StreamReader streamReader = null, 
+            ChannelWriter<ISqServerEvent> sqEventsChannel = null,
+            ISqServerSentEventParser sqServerSentEventParser = null,
+            CancellationToken? token = null)
+        {
+            token ??= CancellationToken.None;;
```
georgii-borovinskikh-sonarsource, line 141: do the same with every argument?

SonarQube.Client.Tests/Models/ServerSentEvents/SSEStreamWriterTests.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            var testSubject = CreateTestSubject(streamReader, channel);
+
+            await testSubject.BeginListening();
+
+            streamReader.EndOfStream.Should().BeTrue();
+            channel.Reader.Count.Should().Be(0);
+        }
+
+        [TestMethod, Timeout(10000)]
```
georgii-borovinskikh-sonarsource, line 51: why Timeout?

georgii-borovinskikh-sonarsource (changes requested):

SonarQube.Client/Models/ServerSentEvents/SSEStreamWriter.cs
rita-sonarsource: why?

SonarQube.Client.Tests/Models/ServerSentEvents/SSEStreamWriterTests.cs
rita-sonarsource: What do you mean?

SonarQube.Client.Tests/Models/ServerSentEvents/SSEStreamWriterTests.cs
rita-sonarsource: In case there is a bug, and the async hangs. So this way the test will terminate

SonarQube.Client/Models/ServerSentEvents/SSEStreamWriter.cs
rita-sonarsource: For some reason the test passes now. Hmm.

SonarQube.Client/Models/ServerSentEvents/SSEStreamWriter.cs
```diff
@@ -35,39 +36,60 @@ public interface ISSEStreamWriter : IDisposable
...
+                {
+                    var parsedEvent = sqServerSentEventParser.Parse(eventLines);
+                    await sqEventsChannel.WriteAsync(parsedEvent, cancellationToken);
+                    eventLines.Clear();
+                }
+                else
+                {
+                    eventLines.Add(line);
+                }
+            }
```
duncanp-sonar, line 87: What will happen in the event of an exception? (e.g. user loses internet connection briefly)

What is responsible for notification and reconnecting?

SonarQube.Client/Models/ServerSentEvents/SSEStreamWriter.cs
```diff
@@ -35,39 +36,60 @@ public interface ISSEStreamWriter : IDisposable
...
+            var eventLines = new List<string>();
+
+            while (!networkStreamReader.EndOfStream && !cancellationToken.IsCancellationRequested)
+            {
+                var line = await networkStreamReader.ReadLineAsync();
+                var isEventEnd = string.IsNullOrEmpty(line);
+
+                if (isEventEnd)
+                {
+                    var parsedEvent = sqServerSentEventParser.Parse(eventLines);
```
duncanp-sonar, line 79: Will the parser ever return `null`? (e.g. an unrecognised event)

Could the parser throw? If so, that is something we could recover from here since the stream reader is still in a known state i.e. we've read a series of related lines that comprise a logical event. If there was a problem parsing the event we could just ignore it an continue to read the next event.

duncanp-sonar (commented):
Looks good. A couple of questions about handling and recovering from failures.

SonarQube.Client/Models/ServerSentEvents/SSEStreamWriter.cs
rita-sonarsource: You're right, I missed this. Opened separate tasks.

SonarQube.Client/Models/ServerSentEvents/SSEStreamWriter.cs
rita-sonarsource: The parser should not throw but you're right it can return null, I'll fix

PR #152, opened 2023-02-03
Support Taint server events

rita-sonarsource:


SonarQube.Client.Tests/Models/ServerSentEvents/SSEStreamReaderTests.cs
```diff
@@ -137,6 +137,47 @@ public async Task GetNextEventOrNullAsync_IssueChangedEventType_DeserializedEven
...
+            result.Should().BeEquivalentTo(
+                new TaintVulnerabilityClosedServerEvent(
+                    projectKey: "projectKey1",
+                    key: "taintKey"));
+        }
+
+        [TestMethod]
+        public async Task GetNextEventOrNullAsync_TaintVulnerabilityRaisedEventType_DeserializedEvent()
+        {
+            const string serializedTaintVulnerabilityRaisedEvent =
```
rita-sonarsource, line 163: We actually get the entire payload back, and seems like it comes with the secondary location hashes too: https://github.com/SonarSource/sonarlint-core/blob/171ca4d75c24033e115a81bd7481427cd1f39f4c/server-api/src/test/java/org/sonarsource/sonarlint/core/serverapi/push/PushApiTests.java#L354

So we might not need to pull anything on SLVS side and could instead just deserialize the data from the event. I asked Damien about it but he's off today. For now I deserialized the other fields, will revisit it next week.

PR #153, opened 2023-02-07
Support retries for recoverable http responses

rita-gorokhod:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3648


SonarQube.Client/Api/V9_4/GetSonarLintEventStream.cs
```diff
@@ -42,6 +46,42 @@ internal class GetSonarLintEventStream : RequestBase<Stream>, IGetSonarLintEvent
...
+            {
+                do
+                {
+                    requestAttempts++;
+
+                    var result = await InvokeUncheckedAsync(httpClient, token);
+
+                    if (ShouldRetry(result) && requestAttempts < MaxNumberOfRequestAttempts)
+                    {
+                        continue;
```
rita-sonarsource, line 66: On the java side they also add a delay. I asked Damien if that was part of the API agreement or an improvement.

duncanp-sonar (commented):
The structure looks ok, but I think it needs some kind of delay otherwise it won't be much use in practice.

Alternatively: if we are also implementing a periodic batch update mechanism, could we do something simpler there? e.g. make the batch update open a new streaming connection each time?

SonarQube.Client/Api/V9_4/GetSonarLintEventStream.cs
georgii-borovinskikh-sonarsource: Yes, it does need a delay, preferably an exponentially increasing one. The reason is, if the server is struggling under load, we should not try to reconnect immediately

SonarQube.Client/Api/V9_4/GetSonarLintEventStream.cs
```diff
@@ -42,6 +46,42 @@ internal class GetSonarLintEventStream : RequestBase<Stream>, IGetSonarLintEvent
...
+                    result.EnsureSuccess();
+
+                    return result.Value;
+                } while (true);
+            }
+        }
+
+        private static bool ShouldRetry(Result<Stream> result)
+        {
+            var isRecoverableStatus = result.StatusCode != HttpStatusCode.Forbidden &&
```
georgii-borovinskikh-sonarsource, line 78: Make this an extension method maybe? It would not add much value in terms of reusing the code, but it would look nicer

SonarQube.Client/Api/V9_4/GetSonarLintEventStream.cs
```diff
@@ -42,6 +46,42 @@ internal class GetSonarLintEventStream : RequestBase<Stream>, IGetSonarLintEvent
...
+
+                    if (ShouldRetry(result) && requestAttempts < MaxNumberOfRequestAttempts)
+                    {
+                        continue;
+                    }
+
+                    result.EnsureSuccess();
+
+                    return result.Value;
+                } while (true);
```
georgii-borovinskikh-sonarsource, line 72: A suggestion, but not a strong one: maybe we can move the exit condition to the while statement argument instead of `true`, it can be easier to read in some cases.

SonarQube.Client/Api/V9_4/GetSonarLintEventStream.cs
rita-sonarsource: Hmm, it will need to be private since it's only unique to this API, I don't see us reusing it anywhere.

duncanp-sonar: Extension methods can't be private or in nested classes, so we couldn't really hide it properly. I don't think it would be an improvement overall.

PR #154, opened 2023-02-07
Remove dependency on Channels and BeginListening() method

rita-sonarsource:
Simplifying the API from SLVS' point of view. No more BeginListening() method, SLVS needs to call only ReadAsync.

Renamed SSEStreamWriter to SqSSEStreamReader, and SSEStream to SSEStreamReader.
SSEStreamReader now directly depends on SqSSEStreamReader, without the channel layer in the middle.


SonarQube.Client.Tests/Models/ServerSentEvents/SSEStreamReaderTests.cs
```diff
@@ -37,26 +35,12 @@ namespace SonarQube.Client.Tests.Models.ServerSentEvents
     [TestClass]
     public class SSEStreamReaderTests
     {
-        [TestMethod]
```
rita-sonarsource, line 40: SSEStreamReader doesn't accept a token anymore so the test is not needed

SonarQube.Client.Tests/Models/ServerSentEvents/SqSSEStreamReaderTests.cs
```diff
@@ -146,36 +137,17 @@ public void Dispose_ClosesStream()
             networkStreamReader.BaseStream.Should().BeNull();
         }
 
-        [TestMethod]
```
rita-sonarsource, line 149: Not relevant anymore

SonarQube.Client/Models/ServerSentEvents/ServerContract/SqSSEStreamReader.cs
```diff
@@ -21,51 +21,46 @@
...
             ISqServerSentEventParser sqServerSentEventParser)
         {
             this.networkStreamReader = networkStreamReader;
             this.cancellationToken = cancellationToken;
-            this.sqEventsChannel = sqEventsChannel;
             this.sqServerSentEventParser = sqServerSentEventParser;
         }
 
-        public async Task BeginListening()
+        public async Task<ISqServerEvent> ReadAsync()
```
duncanp-sonar, line 63: The general pattern for `async` APIs is to pass the cancellation token in the method, which makes passing it in the constructor look a bit odd; it isn't clear from the interface that work can be cancelled.

I imagine there are more complex scenarios where the cancellation token used by the caller might change between calls, but it probably doesn't make much difference in our case.

duncanp-sonar (approved):
Nice - much simpler.

SonarQube.Client/Models/ServerSentEvents/ServerContract/SqSSEStreamReader.cs
georgii-borovinskikh-sonarsource: I think it's because initially I suggested passing it to the constructor + it didn't make much sense otherwise while we had BeginListening that also needed to be cancelled, but now I think this is a good suggestion to change to the regular pattern

rita-sonarsource: It has to be the same cancellation token that is being passed by SLVS, and SLVS passes it when it calls SonarQubeClient.CreateSSEStreamReader. I'm not sure it makes sense to pass it again, and in fact it might lead to a bug that SLVS can pass something else accidentally.

georgii-borovinskikh-sonarsource: The token is created by SSESession and ReadAsync is also called by it, so maybe just pass it directly to ReadAsync each time it gets called?

SonarQube.Client/Models/ServerSentEvents/SSEStreamReader.cs
```diff
@@ -18,26 +18,25 @@
...
     {
         /// <summary>
+        /// Wraps the stream response from the server, reads from it and converts it to <see cref="IServerEvent"/>.
         /// Will block the calling thread until an event exists or the connection is closed.
-        /// Can throw an exception if the event is not a valid <see cref="IServerEvent"/>
         /// </summary>
         /// <returns>
-        /// Will return null if the underlying event is unsupported or if there was a problem parsing it.
+        /// Can return null (i.e. if the underlying event type is unsupported).
+        /// Will throw if there was a problem reading from the underlying stream.
```
georgii-borovinskikh-sonarsource, line 39: Is this true?

SonarQube.Client/Models/ServerSentEvents/ServerContract/SqSSEStreamReader.cs
rita-sonarsource: But then it's possible to pass a different token in ReadAsync than the one that was passed originally to CreateSSEStreamReader. Would we want that? I think it can be a dangerous side effect for not much value.

rita-sonarsource: Spoke with @georgii-borovinskikh-sonarsource offline: the same token that is passed to Create is then used everywhere, including in the http client. So it's better to keep the same one and not allow SLVS to pass another one by mistake.

georgii-borovinskikh-sonarsource (approved):

SonarQube.Client/Models/ServerSentEvents/SSEStreamReader.cs
rita-sonarsource: Also noticed it, right now no, but I think it should. It's what we talked about yesterday, I think SLVS will need to handle it and make a new session.
I'll remove the comment for now but I think we should throw it.

georgii-borovinskikh-sonarsource: Comment out the comment :-P

PR #155, opened 2023-02-07
Add taint fields to TaintVulnerabilityRaisedServerEvent

rita-sonarsource:
Fixes https://github.com/SonarSource/sonarlint-visualstudio/issues/3654


SonarQube.Client.Tests/Models/ServerSentEvents/TaintVulnerabilityRaisedServerEventTests.cs
```diff
@@ -31,25 +33,126 @@ public class TaintVulnerabilityRaisedServerEventTests
...
+						""endLineOffset"": 3,
+						""hash"": ""hash2""
+					}
+				}
+			]
+		}
+	]
+}";
+
+            var result = JsonConvert.DeserializeObject<TaintVulnerabilityRaisedServerEvent>(serializedTaintVulnerabilityRaisedEvent);
```
rita-sonarsource, line 109: The assertions are disgusting but I don't have a better solution

SonarQube.Client/Models/ServerSentEvents/ClientContract/ITaintIssue.cs
```diff
@@ -0,0 +1,121 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using System;
+
+namespace SonarQube.Client.Models.ServerSentEvents.ClientContract
+{
+    public interface ITaintIssue
+    {
+        //DateTimeOffset CreationDate { get; }
```
rita-sonarsource, line 27: for some reason I can't deserialize it, but I don't think we need it anyway

SonarQube.Client/Models/ServerSentEvents/ClientContract/ITaintIssue.cs
```diff
@@ -0,0 +1,121 @@
...
+    public interface ITextRange
+    {
+        int StartLine { get; }
+        int StartLineOffset { get; }
+        int EndLine { get; }
+        int EndLineOffset { get; }
+        string Hash { get; }
+    }
+
+    internal class TaintIssue : ITaintIssue
```
rita-sonarsource, line 56: Sadly the old api returns SonarQubeIssue and the new API for taint returns this, and they're similar but not the same.

SonarQube.Client/Models/ServerSentEvents/ClientContract/ITaintServerEvent.cs
```diff
@@ -34,20 +34,26 @@ public interface ITaintServerEvent : IServerEvent
...
+        ITaintIssue Issue { get; }
     }
 
-    public class TaintVulnerabilityRaisedServerEvent : ITaintVulnerabilityRaisedServerEvent
+    internal class TaintVulnerabilityRaisedServerEvent : ITaintVulnerabilityRaisedServerEvent
     {
-        public TaintVulnerabilityRaisedServerEvent(string projectKey, string key, string branch)
+        public TaintVulnerabilityRaisedServerEvent(string projectKey, string key, string branch, 
+            //DateTimeOffset creationDate, 
+            string ruleKey, SonarQubeIssueSeverity severity, SonarQubeIssueType type,
```
rita-sonarsource, line 44: The original json has the fields in a flat structure, but there is also a separate new taint API (that we should use) and it returns the new structure.
I'm not sure if it's better to have the TaintVulnerabilityRaisedServerEvent be identical to the json (flat, and inheriting from ITaintIssue), or the way I did it (which requires a bit of a "hack" to convert the flat fields into the issue).

LMK what you think.

SonarQube.Client.Tests/Models/ServerSentEvents/TaintVulnerabilityRaisedServerEventTests.cs
georgii-borovinskikh-sonarsource: Maybe like this?
```
var deserialized = JsonConvert.DeserializeObject<T>(JsonConvert.SerializeObject<T>(original));
deserialized.Should().BeEquivalentTo(original);
```

SonarQube.Client/Models/ServerSentEvents/ClientContract/ITaintIssue.cs
georgii-borovinskikh-sonarsource: Need a custom converter maybe?

SonarQube.Client/Models/ServerSentEvents/ClientContract/ITaintIssue.cs
georgii-borovinskikh-sonarsource: Is this a big or a small issue? I don't have the needed context to understand this :-(

SonarQube.Client/Models/ServerSentEvents/ClientContract/ITaintServerEvent.cs
georgii-borovinskikh-sonarsource: Need more context, don't know what is best here

SonarQube.Client/Models/ServerSentEvents/ClientContract/ITaintIssue.cs
rita-sonarsource: Could be, it works out of the box in our other objects, maybe the format is different on Sq side. Either way, I don't think we need it for now so maybe I'll just not bother with it

SonarQube.Client/Models/ServerSentEvents/ClientContract/ITaintIssue.cs
rita-sonarsource: I was hoping that I could simply use the existing model SonarQubeIssue but sadly I can't, so I had to create a new one

SonarQube.Client/Models/ServerSentEvents/ClientContract/ITaintServerEvent.cs
rita-sonarsource: Let's see what @duncanp-sonar says

duncanp-sonar: As discussed offline, it should make too much difference since we can change it later if necessary.
The consensus was that the nested version would be nicer to work with in SLVS so we'd start with that.

PR #161, opened 2023-02-17
Draft: add exception handling and reconnections to SqSSEStreamReader

georgii-borovinskikh-sonarsource:


SonarQube.Client/Models/ServerSentEvents/ServerContract/SqSSEStreamReader.cs
```diff
@@ -43,31 +43,46 @@ internal interface ISqSSEStreamReader : IDisposable
...
         }
 
         public async Task<ISqServerEvent> ReadAsync()
         {
             var eventLines = new List<string>();
 
-            while (!networkStreamReader.EndOfStream && !cancellationToken.IsCancellationRequested)
+            while (!disposalToken.IsCancellationRequested
+                   && await CheckCurrentStreamReaderAndReconnect() 
+                   && !currentNetworkStreamReader.EndOfStream /*I recently discovered that this tries to read the underlying stream synchronously*/)
```
georgii-borovinskikh-sonarsource, line 77: oh, and one more thing - .EndOfStream can also throw, so we definitely need to get rid of it

PR #162, opened 2023-02-17
Make SSEStreamReader not suppress underlying stream errors

georgii-borovinskikh-sonarsource:


SonarQube.Client/Models/ServerSentEvents/SSEStreamReader.cs
```diff
@@ -92,19 +92,13 @@ private async Task<ISqServerEvent> ReadNextEventAsync()
...
-            {
-                return null;
+                return await sqSSEStreamReader.ReadAsync();
             }
             catch (Exception ex)
             {
-                logger.Debug($"[SSEStreamReader] Failed to read sq event: {ex}");
-
-                return null;
+                logger.Warning($"[SSEStreamReader] Underlying stream failure while reading next event: {ex}");
```
duncanp-sonar, line 99: Logging: we need to co-ordinate the logging between the web client and SLVS so we don't overwhelm the users with warnings (less important for debug messages since they are off by default). Just something to bear in mind when we make the changes on the SLVS side.

duncanp-sonar (approved):

SonarQube.Client/Models/ServerSentEvents/SSEStreamReader.cs
georgii-borovinskikh-sonarsource: I felt like we should log this closer to the source. We still can log it on slvs side as well, maybe in less detail or in debug instead

SonarQube.Client/Models/ServerSentEvents/SSEStreamReader.cs
```diff
@@ -92,19 +92,13 @@ private async Task<ISqServerEvent> ReadNextEventAsync()
...
+                sqSSEStreamReader.Dispose();
```
rita-sonarsource, line 100: I feel this is not done in the right place.
SSESession already disposes itself as a response to this message, so perhaps it should dispose `sseStreamReader`, that will dispose everything else. WDYT?

SonarQube.Client/Models/ServerSentEvents/SSEStreamReader.cs
rita-sonarsource: IMO it's better not to do anything here and to let the errors propagate to SLVS.
SLVS will then log, cancel and dispose everything.

rita-sonarsource (changes requested):
I think we should throw everything and not log it/not dispose.

SonarQube.Client/Models/ServerSentEvents/SSEStreamReader.cs
georgii-borovinskikh-sonarsource: I feel like, from a library design POV, this should happen here. Disposing it on SLVS side works only because we own both the library and the code that uses it. I think we should keep the knowledge about the internal components private. That doesn't mean we don't need to dispose it on SLVS side, it just disposes different things: one cancels the token in case of both regular and exceptional cases, the other disposes the network stream only in the case of exceptional failures. Plus, our Dispose methods are idempotent, so nothing bad is going to happen.

SonarQube.Client/Models/ServerSentEvents/SSEStreamReader.cs
georgii-borovinskikh-sonarsource: It's a transport layer error and I feel like it should be logged at the transport layer. Otherwise, what's the point of having the Error/Warn logger in SQS at all, if the only purpose of it is to write Debug logs?

rita-sonarsource: You should enable our debug logs and try to see what happens... ;) we already log way too much, adding another log won't provide any additional information and will just clutter it further

SonarQube.Client/Models/ServerSentEvents/SSEStreamReader.cs
duncanp-sonar: `SSEStreamReader` isn't currently disposable, which is perhaps a bit odd.
However, that means at the moment this is the only place we can dispose this class.
I would leave it as it is for now and move on.

rita-sonarsource: Yes, it's leftovers from when reader and writer were separate and only the writer was disposable. When I changed the reader to hold the writer, I missed the dispose 🙈

